PR #24, opened 2020-03-05
Rework Exp(), remove ExpF()

chfast:
This saves one and only allocation in Exp().

Is `ExpF()` needed as a public function?


chfast:
This is almost done, but I realized that arguments can alias themselves and I need to add unit test for this and fix `Exp()` implementation.

chfast:
> This is almost done, but I realized that arguments can alias themselves and I need to add unit test for this and fix `Exp()` implementation.

Fixed.

PR #27, opened 2020-03-11
Mul: Use bits.Mul64()

chfast:
`Mul()` is now 3x faster.

### DIFF
--- a/uint256.go
+++ b/uint256.go
@@ -179,8 +179,6 @@ func (z *Int) Clone() *Int {
 	return &Int{z[0], z[1], z[2], z[3]}
 }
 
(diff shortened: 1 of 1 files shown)


chfast:
The finding here is that `math/bits` has an intrinsic that is exactly what `mulInto*128` were implementing. On x86_64 this is single instruction `MUL` 64x64 -> 128 instead of the procedure composed of 32x32 -> 64 multiplications.

In https://github.com/chfast/intx the speedup is not so dramatic, but maybe some other effect on the optimizer took place.

Similarly, we can use `bits.Add64()` and `bits.Sub64()` to simplify addition/subtraction.

PR #30, opened 2020-03-13
Add "full" 256x256->512 multiplication

chfast:
This is used for `MulMod()` so `big.Mul()` is not needed any more (we still need `big.Mod()`).
This makes `small/uint256-6` case slower because we also removed the shortcut for `128x128` case.
The same procedure can be used to implement `Mul()`, but when `umul()` is used directly the final result is slower - optimizer is not clever enough to notice that the top 4 words in `umul()` are not used in the end.
Inner loop unroll removes any bounds checking.
```
_MulMod/large/big-6       384ns ± 1%   382ns ± 0%   -0.47%  (p=0.048 n=5+5)
_MulMod/large/uint256-6   802ns ± 0%   721ns ± 0%  -10.05%  (p=0.000 n=5+4)
_MulMod/small/big-6      93.9ns ± 0%  93.6ns ± 0%     ~     (p=0.103 n=5+5)
_MulMod/small/uint256-6  33.5ns ± 0%  41.1ns ± 0%  +22.69%  (p=0.016 n=4+5)
```

### DIFF
--- a/uint256.go
+++ b/uint256.go
@@ -269,6 +269,24 @@ func (z *Int) Sub(x, y *Int) {
 	z.SubOverflow(x, y) // Inlined.
 }
 
+// umul computes full 256 x 256 -> 512 multiplication.
+func umul(x, y *Int) [8]uint64 {
+	var res [8]uint64
+	for j := 0; j < len(y); j++ {
+		var carry uint64
+		for i := 0; i < len(x); i++ {
+			var p [2]uint64
+			p[1], p[0] = bits.Mul64(x[i], y[j])
+			addTo128(p[:], res[i+j], 0)
+			addTo128(p[:], carry, 0)
+			res[i+j] = p[0]
+			carry = p[1]
+		}
+		res[j+len(x)] = carry
+	}
+	return res
+}
+
 // Mul sets z to the sum x*y
 func (z *Int) Mul(x, y *Int) {
 
--- a/uint256.go
+++ b/uint256.go
@@ -269,18 +269,23 @@ func (z *Int) Sub(x, y *Int) {
 	z.SubOverflow(x, y) // Inlined.
 }
 
+// umulStep computes (carry, z) = z + (x * y) + carry.
+func umulStep(z, x, y, carry uint64) (uint64, uint64) {
+	ph, p := bits.Mul64(x, y)
+	p, carry = bits.Add64(p, carry, 0)
+	carry, _ = bits.Add64(ph, 0, carry)
+	p, carry1 := bits.Add64(p, z, 0)
+	carry, _ = bits.Add64(carry, 0, carry1)
+	return p, carry
+}
+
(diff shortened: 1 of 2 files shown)


holiman:
Nice!
I pushed a commit on top, which adds a new conversion function from `*big.Int` to `uint256.Int`. If I apply that :
```diff
diff --git a/uint256.go b/uint256.go
index 63f3d4a..4ca0df9 100644
--- a/uint256.go
+++ b/uint256.go
@@ -618,7 +618,7 @@ func (z *Int) MulMod(x, y, m *Int) *Int {
        // if they fit within 256 bytes. But for now just wrap big.Int instead
        bp := new(big.Int)
        bp.SetBytes(pbytes[:])
-       z.SetFromBig(bp.Mod(bp, m.ToBig()))
+       z.SetFromBigFast(bp.Mod(bp, m.ToBig()))
        return z
 }
```
Then it's possible to further improve the worst case with 20% 
```
name                     old time/op    new time/op    delta
_MulMod/large/big-6         531ns ±21%     539ns ±18%     ~     (p=0.421 n=5+5)
_MulMod/large/uint256-6    1.08µs ±12%    0.86µs ± 0%  -20.57%  (p=0.016 n=5+4)
_MulMod/small/big-6         129ns ±17%     125ns ± 4%     ~     (p=0.849 n=5+5)
_MulMod/small/uint256-6    51.4ns ± 0%    52.3ns ± 5%     ~     (p=1.000 n=5+5)

name                     old alloc/op   new alloc/op   delta
_MulMod/large/big-6          176B ± 0%      176B ± 0%     ~     (all equal)
_MulMod/large/uint256-6      336B ± 0%      304B ± 0%   -9.52%  (p=0.008 n=5+5)
_MulMod/small/big-6         56.0B ± 0%     56.0B ± 0%     ~     (all equal)
_MulMod/small/uint256-6     0.00B          0.00B          ~     (all equal)

name                     old allocs/op  new allocs/op  delta
_MulMod/large/big-6          2.00 ± 0%      2.00 ± 0%     ~     (all equal)
_MulMod/large/uint256-6      6.00 ± 0%      5.00 ± 0%  -16.67%  (p=0.008 n=5+5)
_MulMod/small/big-6          2.00 ± 0%      2.00 ± 0%     ~     (all equal)
_MulMod/small/uint256-6      0.00           0.00          ~     (all equal)
```

I didn't commit actually using that function, because it's not yet complete - I haven't implemented support for 32-bit platforms. 

Tangent: is it possible/easy to enable 32-bit platform testing on CircleCI?

chfast:
Hey, I moved your commit to `conversion` branch. I'd like to take a look there later. In the end the conversion from/to `math.big` will not affect performance in `MulMod` as we will have our own division procedure.

chfast:
> Tangent: is it possible/easy to enable 32-bit platform testing on CircleCI?

Done in https://github.com/holiman/uint256/pull/31, but currently fails in a conversion test.

PR #31, opened 2020-03-14
ci: Test 32-bit version (GOARCH: 386)

chfast:
This tests 32-bit build (GOARCH: 386).

We can also add some big-endian architectures, but this requires using qemu.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,16 +1,28 @@
 version: 2.1
 
+commands:
+  test:
+    parameters:
+      arch:
+        default: "amd64"
+        description: The target architecture.
+        type: enum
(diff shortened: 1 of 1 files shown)


chfast:
The "Test (386)" currently fails because of a conversion test. It should be fixed first before merging.

holiman:
Awesome, thanks! 
Let me know if you want to handle the conversion, otherwise I'll fix it myself

chfast:
> Let me know if you want to handle the conversion, otherwise I'll fix it myself

Yes, I can handle that. Just not sure when, probably on Monday.

chfast:
Included in #32.

PR #34, opened 2020-03-16
Cleanups and optimization of ToBig()

chfast:
```
ToBig/1word-6            64.9ns ± 0%  50.0ns ± 1%  -22.90%  (p=0.008 n=5+5)
ToBig/2words-6           90.6ns ± 0%  50.0ns ± 0%  -44.86%  (p=0.008 n=5+5)
ToBig/3words-6            111ns ± 0%    50ns ± 2%  -55.14%  (p=0.008 n=5+5)
ToBig/4words-6            126ns ± 0%    50ns ± 1%  -60.59%  (p=0.008 n=5+5)
_MulMod/large/uint256-6   628ns ± 0%   531ns ± 0%  -15.45%  (p=0.016 n=5+4)
```

1. Merge FromBig() and NewFromBig(). They were identical.
2. Adds more unit tests. Moves conversion tests to `conversion_test.go`.
3. Adds a benchmark for `ToBig()`.
4. Optimizes `ToBig()` in a similar way as `FromBig()` although in this case implementation is much simpler.

Probably the best to review commit by commit.

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -16,8 +16,9 @@ const (
 	maxWords = 256 / bits.UintSize // number of big.Words in 256-bit
 )
 
-// NewFromBig creates new Int from big.Int.
-func NewFromBig(b *big.Int) (*Int, bool) {
+// FromBig is a convenience-constructor from big.Int.
+// Returns a new Int and whether overflow occurred.
+func FromBig(b *big.Int) (*Int, bool) {
 	z := &Int{}
 	overflow := z.SetFromBig(b)
 	return z, overflow
--- a/conversion.go
+++ b/conversion.go
@@ -16,11 +16,25 @@ const (
 	maxWords = 256 / bits.UintSize // number of big.Words in 256-bit
 )
 
+// ToBig returns a big.Int version of z.
 func (z *Int) ToBig() *big.Int {
-	x := new(big.Int)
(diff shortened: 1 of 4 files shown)


conversion.go
```diff
@@ -16,8 +16,29 @@ const (
 	maxWords = 256 / bits.UintSize // number of big.Words in 256-bit
 )
 
-// NewFromBig creates new Int from big.Int.
-func NewFromBig(b *big.Int) (*Int, bool) {
+func (z *Int) ToBig() *big.Int {
```
holiman:
```suggestion
// ToBig returns a big.Int version of z
func (z *Int) ToBig() *big.Int {
```

conversion_test.go
```diff
@@ -0,0 +1,151 @@
+// Copyright 2020 Martin Holst Swende. All rights reserved.
```
holiman: I guess we could remove my explicit name, and use AUTHORS instead. I kept it in my name originally, so I would be sure that there would be no problems to change the license at a later point. I guess we're good with the current license?

chfast: We can handle that separately some other time. I this case I copied code from another file so I used the same one.

I'm fine with the license. My default one is Apache2, but I rather don't care witch open-source license it is.

PR #38, opened 2020-03-17
test: Fuzz with go-fuzz

chfast:
1. Adds basic go-fuzz fuzzer - can be extended later.
2. Collects testing code coverage (both amd64 and 386) and uploads to codecov for visual report.
3. The fuzzer runs on CI for one minute. The corpus is preserved.

### DIFF
--- a/fuzz.go
+++ b/fuzz.go
@@ -1,0 +1,58 @@
+// +build gofuzz
+
+package uint256
+
+const (
+	opUdivrem = 0
+	opMul     = 1
+)
+
+func Fuzz(data []byte) int {
+	if len(data) != 65 {
+		return 0
+	}
+
+	op := data[0]
+
+	var x, y Int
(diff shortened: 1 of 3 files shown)


fuzz.go
```diff
@@ -0,0 +1,58 @@
...
+			panic("invalid quotient")
+		}
+
+		if !r.Eq(er) {
+			panic("invalid remainder")
+		}
+
+	case opMul:
+		var p Int
+		p.Mul(&x, &y)
```
holiman: Two other cases that would be interesting to cover (in general, for any test like this): 
- `p.Mul(&x, &y)` where `p` is non-zero, 
- `x.Mul(&x, &y)` and `y.Mul(&x, &y)`
And lastly, to verify that no (non-receiver) parameters were modified during the op

chfast: Ok, I will extend it in the next PR.

PR #39, opened 2020-03-18
div: Faster implementation of reciprocal2by1()

chfast:


holiman:
Abandoned?

chfast:
I noticed that on the recent Intel hardware the "native" use of `div` instruction is faster than the "software" implementation. Finally.

PR #46, opened 2020-04-22
test: Create generic test runner for binary ops

chfast:
This allows test any binary op using predefined test cases. In preparation for testing bitwise ops.


chfast:
> LGTM. One thing to consider: should we in this generic executor also check for "op does not modify input parameters"?

Yes. Originally I wanted to have separate set of tests for this, but once we have generic runner I think it is good place for it.

holiman:
:+1:

PR #49, opened 2020-04-28
test: Add TestTernOp() for {Add,Mul}Mod ops

chfast:
All this to test the case where addition does not overflow in `AddMod()` and regular `Mod()` can be used.

https://codecov.io/gh/holiman/uint256/compare/master...tests/changes#L179


holiman:
One more line of code coverage!! :partying_face: :tada: 

(but seriously though: it's a lot better to structure the tests this way, so great work!)

Issue #50, opened 2020-04-28
1.0 roadmap


chfast:
1. [x] Full test coverage: https://codecov.io/gh/holiman/uint256
2. [x] Inspect `Exp()` implementation (constant time, but gas cost proportional to exponent size).
3. [x] Remove unneeded public functions.
4. [x] Fix `SDiv()`, `SMod()` - they modify arguments.
5. [x] Review decision to return zero on division by zero. #62 #64
6. [x] ~~Test arm architecture?~~
7. [x] ~~Test any big-endian architecture?~~ #60
8. [x] Unify names: `Sub64()` vs `LtUint64()`. #65
9. [x] Run benchmarks against 0.1 version.

chfast:
Comparison with `v0.1.0`:
```
name                      old time/op  new time/op  delta
_Add/single/uint256-6     2.00ns ± 0%  2.00ns ± 0%     ~     (all equal)
_Sub/single/uint256-6     2.00ns ± 0%  2.00ns ± 0%     ~     (all equal)
_Sub/single/uint256_of-6  2.00ns ± 0%  2.00ns ± 0%     ~     (all equal)
Mul/single/uint256-6      15.0ns ± 0%   8.7ns ± 0%  -41.75%  (p=0.000 n=10+10)
Square/single/uint256-6   16.1ns ± 0%   6.3ns ± 0%  -60.68%  (p=0.000 n=10+10)
_And/single/uint256-6     1.77ns ± 1%  1.76ns ± 1%     ~     (p=0.070 n=10+10)
_Or/single/uint256-6      1.77ns ± 0%  1.77ns ± 0%     ~     (all equal)
_Xor/single/uint256-6     1.77ns ± 0%  1.77ns ± 1%     ~     (p=0.353 n=7+10)
_Cmp/single/uint256-6     2.83ns ± 0%  2.83ns ± 0%     ~     (all equal)
Lt/large/uint256-6        2.78ns ± 0%  2.78ns ± 0%     ~     (p=1.000 n=10+10)
Lt/small/uint256-6        2.78ns ± 0%  2.78ns ± 0%     ~     (all equal)
_Lsh/n_eq_0/uint256-6     4.50ns ± 0%  4.50ns ± 0%     ~     (all equal)
_Lsh/n_gt_192/uint256-6   4.75ns ± 0%  4.75ns ± 0%     ~     (all equal)
[...]

PR #52, opened 2020-04-29
Changes to SDiv() and SMod()

chfast:
This clean ups and fixes issues in tests.
It also modifies the implementation not to modify arguments.
Method are renamed:
- `Sdiv()` -> `SDiv()`
- `Smod()` -> `SMod()`
I'm not sure these are substantially better names but at least my editor is not complaining any more.


chfast:
They are still allocation free. I'm not checking performance changes on regular basis, but I plan to check benchmarks when done with API and test changes. And compare that with 0.1. The 1.0 release plan is being tracked here: #50.

PR #53, opened 2020-04-30
Changes and tweaks to increase test coverage

chfast:


chfast:
Looks like `SignExtend()` still needs a fix...

uint256.go
```diff
@@ -1118,25 +1098,25 @@ func (z *Int) Exp(base, exponent *Int) *Int {
...
 }
 
-//Extend length of two’s complement signed integer
+// SignExtend extends length of two’s complement signed integer,
 // sets z to
-//  - num if back  > 31
+//  - num if back > 31
 //  - num interpreted as a signed number with sign-bit at (back*8+7), extended to the full 256 bits
-func (z *Int) SignExtend(back, num *Int) {
+func (z *Int) SignExtend(num, back *Int) *Int{
```
holiman, line 1105: Dangerous API-change to swap two same-type arguments like that... How come you want to swap them?

chfast: True. Also the arg names are at least confusing, so maybe the implementation was done wrongly. We rather want `(value, byte_number)`, but previously it was the other way around.

Also it may be `(x *Int, n uint)` as in shifts. However, what will require additional check on the EVM side.

chfast: Here you have tests in "natural" order: https://github.com/holiman/uint256/pull/53/files#diff-68d84e97126bc30783bdabd350595432R427-R440.

chfast: I removed the controversial commit from this PR.

holiman: I don't particularly care though, the only ones it'll bite is ourselves at this point :) 
No official 1.0 release yet, so should be OK to change the API

chfast: I will make another PR out of it and probably change arg names for clarity.

chfast:
Can I merge this one?

PR #54, opened 2020-05-04
Remove (*Int).Hex()

chfast:


conversion.go
```diff
@@ -87,3 +88,7 @@ func (z *Int) SetFromBig(b *big.Int) bool {
 	}
 	return overflow
 }
+
```
holiman, line 91: perhaps copy the docstring from big.Int ?

chfast: Done.

PR #55, opened 2020-05-04
Refactor ExtendSign() and add tests

chfast:


uint256.go
```diff
@@ -1098,25 +1098,26 @@ func (z *Int) Exp(base, exponent *Int) *Int {
 	return z.Copy(&res)
 }
 
-//Extend length of two’s complement signed integer
+// SignExtend extends length of two’s complement signed integer,
```
holiman, line 1101: Maybe we can change the name to `ExtendSign` to break the API (since we're swapping the arguments) ?

chfast: Done.

PR #56, opened 2020-05-04
Replace panic() with compilation-time check

chfast:
Replace panic() checking for architecture other than 32 or 64 bit
with compilation-time checks using uint constant underflow trick.


conversion.go
```diff
@@ -15,6 +15,9 @@ import (
 
 const (
 	maxWords = 256 / bits.UintSize // number of big.Words in 256-bit
+
```
holiman, line 18: Would perhaps be in order to put a comment  here to explain what the code below does

PR #60, opened 2020-05-05
ci: Add GOARCH=ppc64 (big-endian)

chfast:


chfast:
I tested this locally and all tests pass.
But it seems difficult to run inside docker - `qemu-ppc64 uint256.test` crashes.

PR #69, opened 2020-05-17
Fix some comments

yperbasis:


uint256.go
```diff
@@ -247,7 +243,7 @@ func umul(x, y *Int) [8]uint64 {
 	return res
 }
 
-// Mul sets z to the sum x*y
+// Mul sets z to the product x*y
```
axic, line 246: `product of x*y`?

holiman, line 246:
```suggestion
// Mul sets z to the product of x*y
```

holiman:
Thanks!

PR #70, opened 2020-06-01
Optimized setbytes-functions

holiman:
This was a little experiment that turned out pretty nicely. 

Go-ethereum branch for experimentation: https://github.com/holiman/go-ethereum/tree/fixedbig_optimizations

Using size-specific setbyte methods, things got a lot faster, `1.34µs` vs `165ns`: 
```
name                          old time/op    new time/op    delta
SetBytes/SetBytes-generic-6     1.34µs ± 1%    1.34µs ± 3%    ~     (p=0.465 n=8+9)
SetBytes/SetBytes-specific-6     178ns ± 5%     165ns ± 3%  -7.39%  (p=0.000 n=10+9)
```

By generating code for PUSHX instructions, I got the following improvements in geth:
```
name           old time/op    new time/op    delta
PushPopLoop-6    209ms ±101%      81ms ± 7%  -61.44%  (p=0.000 n=10+10)

name           old alloc/op   new alloc/op   delta
PushPopLoop-6    42.9kB ± 0%    42.6kB ± 0%   -0.66%  (p=0.000 n=8+10)

name           old allocs/op  new allocs/op  delta
PushPopLoop-6      73.4 ± 2%      71.0 ± 0%   -3.24%  (p=0.000 n=8+8)
```

Further (but unrelated to this PR), by also generating code for SWAPX/DUPX, I got the following: 
```
name           old time/op    new time/op    delta
SwapDupLoop-6    65.6ms ± 5%    58.0ms ± 5%  -11.60%  (p=0.000 n=10+10)

name           old alloc/op   new alloc/op   delta
SwapDupLoop-6    48.5kB ± 0%    42.5kB ± 0%  -12.36%  (p=0.000 n=9+10)

name           old allocs/op  new allocs/op  delta
SwapDupLoop-6      69.0 ± 0%      70.5 ± 1%   +2.17%  (p=0.000 n=9+10)
```


uint256_test.go
```diff
@@ -1184,3 +1190,175 @@ func TestByte32Representation(t *testing.T) {
 		}
 	}
 }
+
+func TestSetBytes(t *testing.T) {
```
chfast, line 1194: This should go to `conversions_test.go`.

uint256_test.go
```diff
@@ -1184,3 +1190,175 @@ func TestByte32Representation(t *testing.T) {
...
+			t.Errorf("testcase %d: got %x exp %x", i, got, exp)
+		}
+	}
+}
+
+func BenchmarkSetBytes(b *testing.B) {
+
+	val := NewInt()
+	bytearr := hex2Bytes("12131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f3031")
+	b.Run("SetBytes-generic", func(b *testing.B) {
```
chfast, line 1288:
```suggestion
	b.Run("generic", func(b *testing.B) {
```

uint256_test.go
```diff
@@ -1184,3 +1190,175 @@ func TestByte32Representation(t *testing.T) {
...
+			val.SetBytes(bytearr[:26])
+			val.SetBytes(bytearr[:27])
+			val.SetBytes(bytearr[:28])
+			val.SetBytes(bytearr[:29])
+			val.SetBytes(bytearr[:20])
+			val.SetBytes(bytearr[:31])
+			val.SetBytes(bytearr[:32])
+		}
+	})
+	b.Run("SetBytes-specific", func(b *testing.B) {
```
chfast, line 1326:
```suggestion
	b.Run("specific", func(b *testing.B) {
```

holiman:
Nice improvement! 

>Helpers like bigEndianUint40 can be implement as combination of binary.BigEndian calls. But Go compiler generates the same assembly in both cases. So we can also get rid of using binary.BigEndian entirely. Any comments?

Not sure -- are you saying we should copy-paste the remaining functions from binary.BigEndian too? I'd rather not, if there's no gain in it. 

And another things, since the "specific" is ~10x faster than `generic", it might make sense to replace the "generic" with a "switch + specific" ?

holiman:
Oh - Using a switch:
```
BenchmarkSetBytes/generic-6         	 4944483	       216 ns/op	       0 B/op	       0 allocs/op
BenchmarkSetBytes/specific-6        	10182189	       118 ns/op	       0 B/op	       0 allocs/op
```

holiman:
Pushed a naive version now, along with a failing testcase as a reminder to fix it :)

chfast:
> Not sure -- are you saying we should copy-paste the remaining functions from binary.BigEndian too? I'd rather not, if there's no gain in it.

Both variants generate the same assembly (https://godbolt.org/z/wJE3K_):
```go
func bigEndianUint40_1(b []byte) uint64 {
	_ = b[4] // bounds check hint to compiler; see golang.org/issue/14808
	return uint64(b[4]) | uint64(b[3])<<8 | uint64(b[2])<<16 | uint64(b[1])<<24 |
		uint64(b[0])<<32
}

func bigEndianUint40_2(b []byte) uint64 {
	_ = b[4] // bounds check hint to compiler; see golang.org/issue/14808
	return uint64(binary.BigEndian.Uint32(b[1:5])) | uint64(b[0])<<32
}
```
So there are 3 options here:
1. Keep it as it is (mixture of `binary.BigEndian` and single byte shifts).
2. Use `binary.BigEndian` also in `bigEndianUintXX` helpers.
3. Add more helpers based on single byte shifts - import of `"encoding/binary"` not needed.

chfast:
> And another things, since the "specific" is ~10x faster than `generic", it might make sense to replace the "generic" with a "switch + specific" ?

Yes, make sense. I actually thought this is already done, but it was only code in `TestSetBytes`...

holiman:
> So there are 3 options here:
>

I don't really have a strong opinion... All else being equal, I think we should take whichever looks cleanest, so that the code is easy to understand and easy for a reviewer to vet.

I personally think `bigEndianUint40_1` is easier to understand than `bigEndianUint40_2`. I don't think we should explicitly _avoid_ `binary` (option 3), since it's a (hopefully) very stable native dependency.

holiman:
Oh right, another thing. In `SetBytes`, which behaviour do we want? 
```golang
	bytearr := hex2Bytes("AAAA12131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f3031BBBB")
	z.SetBytes(bytearr)
	fmt.Printf("%x\n", z) // 1415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f3031bbbb
	z2.SetBytesx(bytearr)
	fmt.Printf("%x\n", z2) // aaaa12131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f
```

The current master uses the last bytes, whereas my poc used the first bytes. Which makes more sense, and why?

holiman:
I guess, if we want to align with `big.Int`, then we should stick with `master`. Then `SetBytes` is identical to `bigInt.SetBytes()` followed by conversion to `uint256.Int`.

uint256.go
```diff
@@ -23,6 +23,7 @@ func NewInt() *Int {
 
 // SetBytes interprets buf as the bytes of a big-endian unsigned
 // integer, sets z to that value, and returns z.
+/*
```
chfast, line 26: Drop it?

holiman: yeah, why not. You or me?

uint256.go
```diff
@@ -111,7 +189,7 @@ func (z *Int) Uint64() uint64 {
 
 // Uint64WithOverflow returns the lower 64-bits of z and bool whether overflow occurred
 func (z *Int) Uint64WithOverflow() (uint64, bool) {
-	return z[0], z[1] != 0 || z[2] != 0 || z[3] != 0
+	return z[0], (z[1] | z[2] | z[3]) != 0
```
holiman, line 192: Either this isn't go-formatted, or the lines at 800 , 805 aren't go-formatted

chfast: I double checked. It looks this is different depending if part of longer boolean expression (`&&`, `||`) or not.

holiman:
@chfast I rebased it, squashed it lightly and removed the junk

chfast:
Merge!

PR #75, opened 2020-06-05
MulOverflow

yperbasis:
I deliberately used full umul for clarity; this operation can be optimised later if need be.


holiman:
I'm fine with merging this though, improvements can always be done later

PR #76, opened 2020-06-05
optimize + benchmark muloverflow

holiman:
Adds a benchmarks and simplifies/improves `MulOverflow` a bit

```
name                          old time/op    new time/op    delta
MulOverflow/single/uint256-6    25.1ns ± 3%    23.0ns ± 1%  -8.22%  (p=0.016 n=5+4)
MulOverflow/single/big-6        77.0ns ±13%    75.7ns ± 4%    ~     (p=0.937 n=5+5)

name                          old alloc/op   new alloc/op   delta
MulOverflow/single/uint256-6     0.00B          0.00B         ~     (all equal)
MulOverflow/single/big-6         0.00B          0.00B         ~     (all equal)

name                          old allocs/op  new allocs/op  delta
MulOverflow/single/uint256-6      0.00           0.00         ~     (all equal)
MulOverflow/single/big-6          0.00           0.00         ~     (all equal)
```

### DIFF
--- a/benchmarks_test.go
+++ b/benchmarks_test.go
@@ -165,6 +165,34 @@ func BenchmarkMul(bench *testing.B) {
 	bench.Run("single/big", benchmarkBig)
 }
 
+func BenchmarkMulOverflow(bench *testing.B) {
+	benchmarkUint256 := func(bench *testing.B) {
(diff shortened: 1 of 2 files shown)


holiman:
Actually, we might need to be careful here, I'm not sure this is correct if the same backing slice is used for input.. is it?

holiman:
> Actually, we might need to be careful here, I'm not sure this is correct if the same backing slice is used for input.. is it?

Nah that's fine, the `res` from umul is crisp and clean

PR #79, opened 2020-07-13
conversion: add various marshalling methods

holiman:
This PR adds some more marshalling methods. If we want to replace `big.Int` and `hexutil.Big` with `uint256.Int` in go-ethereum, these methods are needed. 

```
Benchmark_EncodeHex/large/uint256-6             10843225                97.2 ns/op            80 B/op          1 allocs/op
Benchmark_EncodeHex/large/big-6                  6547092               190 ns/op             139 B/op          2 allocs/op
Benchmark_DecodeHex/large/uint256-6              9983671               118 ns/op              32 B/op          1 allocs/op
Benchmark_DecodeHex/large/big-6                 10242964               128 ns/op              32 B/op          1 allocs/op
```


holiman:
cc @fjl PTAL

PR #85, opened 2021-07-05
Optimize comparison in `Cmp()`

fyrchik:
We can do it with a single comparison. The speed up for the bad case is ~30%.
```
name                   old time/op    new time/op    delta
_Cmp/single/big-8        13.1ns ± 4%    13.1ns ± 2%     ~     (p=0.887 n=10+9)
_Cmp/single/uint256-8    5.40ns ± 3%    4.69ns ± 3%  -13.17%  (p=0.000 n=10+9)

name                   old alloc/op   new alloc/op   delta
_Cmp/single/big-8         0.00B          0.00B          ~     (all equal)
_Cmp/single/uint256-8     0.00B          0.00B          ~     (all equal)

name                   old allocs/op  new allocs/op  delta
_Cmp/single/big-8          0.00           0.00          ~     (all equal)
_Cmp/single/uint256-8      0.00           0.00          ~     (all equal)
```

### DIFF
--- a/benchmarks_test.go
+++ b/benchmarks_test.go
@@ -294,6 +294,7 @@ func benchmark_Cmp_Big(bench *testing.B) {
 	bench.ResetTimer()
 	for i := 0; i < bench.N; i++ {
 		b1.Cmp(b2)
+		b2.Cmp(b1)
 	}
 }
 func benchmark_Cmp_Bit(bench *testing.B) {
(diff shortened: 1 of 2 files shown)


fyrchik:
I am not sure if this is the right way to change benchmark (we get duration of 2 operations), but it seems simpler than making 2 more benchmarks.

holiman (commented):
LGTM, thanks!

PR #86, opened 2021-08-13
Optimised modular arithmetic targeting elliptic curve operations

daosvik:


gballet:
Before:

```
goos: freebsd
goarch: amd64
cpu: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
BenchmarkMulMod
BenchmarkMulMod/small/uint256
BenchmarkMulMod/small/uint256-8         	24091966	        51.95 ns/op
BenchmarkMulMod/small/big
BenchmarkMulMod/small/big-8             	13057272	        93.92 ns/op
BenchmarkMulMod/mod64/uint256
BenchmarkMulMod/mod64/uint256-8         	 6775998	       162.3 ns/op
BenchmarkMulMod/mod64/big
BenchmarkMulMod/mod64/big-8             	 3723324	       324.2 ns/op
BenchmarkMulMod/mod128/uint256
BenchmarkMulMod/mod128/uint256-8        	 4329858	       247.9 ns/op
BenchmarkMulMod/mod128/big
BenchmarkMulMod/mod128/big-8            	 2085337	       693.0 ns/op
BenchmarkMulMod/mod192/uint256
BenchmarkMulMod/mod192/uint256-8        	 3751779	       318.3 ns/op
BenchmarkMulMod/mod192/big
BenchmarkMulMod/mod192/big-8            	 1637002	       760.2 ns/op
BenchmarkMulMod/mod256/uint256
BenchmarkMulMod/mod256/uint256-8        	 3333724	       343.7 ns/op
BenchmarkMulMod/mod256/big
BenchmarkMulMod/mod256/big-8            	 1535760	       763.2 ns/op
PASS
ok  	github.com/holiman/uint256	15.822s
```

After:

```
goos: freebsd
goarch: amd64
cpu: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
BenchmarkMulMod
BenchmarkMulMod/small/uint256
BenchmarkMulMod/small/uint256-8         	25852512	        46.44 ns/op
BenchmarkMulMod/small/big
BenchmarkMulMod/small/big-8             	13739356	        88.47 ns/op
BenchmarkMulMod/mod64/uint256
BenchmarkMulMod/mod64/uint256-8         	 6403490	       196.1 ns/op
BenchmarkMulMod/mod64/big
BenchmarkMulMod/mod64/big-8             	 3167760	       394.4 ns/op
BenchmarkMulMod/mod128/uint256
BenchmarkMulMod/mod128/uint256-8        	 3329642	       338.7 ns/op
BenchmarkMulMod/mod128/big
BenchmarkMulMod/mod128/big-8            	 1510978	       829.5 ns/op
BenchmarkMulMod/mod192/uint256
BenchmarkMulMod/mod192/uint256-8        	 3204769	       357.8 ns/op
BenchmarkMulMod/mod192/big
BenchmarkMulMod/mod192/big-8            	 1512756	       778.7 ns/op
BenchmarkMulMod/mod256/uint256
BenchmarkMulMod/mod256/uint256-8        	 5124606	       233.6 ns/op
BenchmarkMulMod/mod256/big
BenchmarkMulMod/mod256/big-8            	 1604929	       720.6 ns/op
    benchmarks_test.go:712: Hits/Misses	4788461/1349390
PASS
ok  	github.com/holiman/uint256	17.042s
```

gballet:
Running [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat):

```
(master)> go test -v -run none -bench MulMod -count 5  -benchmem > master.txt
(master)> git checkout opt_modarith
(opt_modarith)> go test -v -run none -bench MulMod -count 5  -benchmem > daosvik.txt
> benchstat master.txt daosvik.txt 
name                     old time/op    new time/op    delta
MulMod/small/uint256-8     46.0ns ±28%    38.7ns ± 2%     ~     (p=0.063 n=5+4)
MulMod/small/big-8          115ns ± 8%     118ns ± 6%     ~     (p=0.460 n=5+5)
MulMod/mod64/uint256-8      174ns ± 8%     172ns ± 6%     ~     (p=1.000 n=5+5)
MulMod/mod64/big-8          332ns ±11%     331ns ± 4%     ~     (p=0.548 n=5+5)
MulMod/mod128/uint256-8     280ns ± 6%     282ns ± 0%     ~     (p=0.730 n=5+4)
MulMod/mod128/big-8         699ns ± 3%     755ns ±12%     ~     (p=0.056 n=5+5)
MulMod/mod192/uint256-8     287ns ± 8%     377ns ±17%  +31.49%  (p=0.008 n=5+5)
MulMod/mod192/big-8         675ns ± 3%     743ns ± 7%  +10.17%  (p=0.016 n=5+5)
MulMod/mod256/uint256-8     288ns ±10%     215ns ±12%  -25.47%  (p=0.008 n=5+5)
MulMod/mod256/big-8         659ns ± 8%     682ns ± 1%     ~     (p=0.190 n=5+4)
```

There seems to be a significant performance hit. These are on a very specific platform, though, I wonder what the results would be on more standard platforms?

daosvik:
Here are my MulMod benchmark results on amd64 and arm64, with go1.13.8 on Ubuntu 20.04.3 LTS.

6-core Ryzen 5 3600 running at 3.6 GHz (boost is off):
```
MulMod/small/uint256-12     30.7ns ± 0%    31.4ns ± 0%   +2.41%  (p=0.016 n=5+4)
MulMod/small/big-12         69.9ns ± 2%    70.5ns ± 1%     ~     (p=0.254 n=5+5)
MulMod/mod64/uint256-12     93.7ns ± 1%    95.6ns ± 0%   +1.99%  (p=0.008 n=5+5)
MulMod/mod64/big-12          247ns ± 1%     248ns ± 1%     ~     (p=0.833 n=5+5)
MulMod/mod128/uint256-12     167ns ± 0%     164ns ± 0%   -1.80%  (p=0.029 n=4+4)
MulMod/mod128/big-12         453ns ± 2%     467ns ± 2%   +2.91%  (p=0.016 n=5+5)
MulMod/mod192/uint256-12     160ns ± 0%     159ns ± 1%   -0.88%  (p=0.048 n=5+5)
MulMod/mod192/big-12         414ns ± 1%     420ns ± 1%     ~     (p=0.063 n=5+5)
MulMod/mod256/uint256-12     155ns ± 0%     108ns ± 0%  -30.32%  (p=0.008 n=5+5)
MulMod/mod256/big-12         397ns ± 2%     406ns ± 2%     ~     (p=0.087 n=5+5)
```

4-core Core i7-1065G7 running at 1.3 GHz:
```
MulMod/small/uint256-8     91.3ns ± 0%    90.2ns ± 6%     ~     (p=0.690 n=5+5)
MulMod/small/big-8          132ns ± 0%     130ns ± 0%   -1.52%  (p=0.000 n=5+4)
MulMod/mod64/uint256-8      212ns ± 0%     205ns ± 0%   -3.67%  (p=0.008 n=5+5)
MulMod/mod64/big-8          337ns ± 0%     337ns ± 0%     ~     (p=1.000 n=5+5)
MulMod/mod128/uint256-8     433ns ± 0%     424ns ± 0%   -2.12%  (p=0.016 n=5+4)
MulMod/mod128/big-8         805ns ± 0%     799ns ± 0%   -0.77%  (p=0.008 n=5+5)
MulMod/mod192/uint256-8     418ns ± 1%     412ns ± 0%   -1.44%  (p=0.016 n=5+4)
MulMod/mod192/big-8         785ns ±14%     752ns ± 0%     ~     (p=0.056 n=5+5)
MulMod/mod256/uint256-8     412ns ± 0%     320ns ±11%  -22.23%  (p=0.016 n=4+5)
MulMod/mod256/big-8         719ns ± 0%     715ns ± 1%     ~     (p=0.127 n=5+5)
```

4-core Cortex-A72 running at 1.5 GHz (Raspberry Pi 4):
```
MulMod/small/uint256-4      186ns ± 0%     185ns ± 0%   -0.54%  (p=0.008 n=5+5)
MulMod/small/big-4          375ns ± 0%     375ns ± 1%     ~     (p=0.841 n=5+5)
MulMod/mod64/uint256-4      416ns ± 0%     413ns ± 0%   -0.86%  (p=0.008 n=5+5)
MulMod/mod64/big-4         1.07µs ± 1%    1.08µs ± 1%     ~     (p=0.095 n=5+5)
MulMod/mod128/uint256-4     741ns ± 0%     736ns ± 0%   -0.78%  (p=0.008 n=5+5)
MulMod/mod128/big-4        2.01µs ± 2%    2.01µs ± 2%     ~     (p=0.690 n=5+5)
MulMod/mod192/uint256-4     746ns ± 0%     742ns ± 0%   -0.56%  (p=0.016 n=5+5)
MulMod/mod192/big-4        1.89µs ± 1%    1.91µs ± 1%   +0.95%  (p=0.008 n=5+5)
MulMod/mod256/uint256-4     721ns ± 0%     612ns ± 0%  -15.17%  (p=0.008 n=5+5)
MulMod/mod256/big-4        1.78µs ± 1%    1.79µs ± 1%     ~     (p=0.690 n=5+5)
```

holiman:
on `cpu: Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz`
```
name                     old time/op    new time/op    delta
MulMod/small/uint256-6     36.8ns ± 3%    36.3ns ± 1%     ~     (p=0.206 n=5+5)
MulMod/small/big-6         70.1ns ±13%    65.9ns ± 4%     ~     (p=0.238 n=5+5)
MulMod/mod64/uint256-6      117ns ± 9%     105ns ± 3%  -10.22%  (p=0.008 n=5+5)
MulMod/mod64/big-6          238ns ± 1%     216ns ± 1%   -9.30%  (p=0.008 n=5+5)
MulMod/mod128/uint256-6     211ns ± 9%     173ns ± 0%  -18.18%  (p=0.016 n=5+4)
MulMod/mod128/big-6         572ns ±11%     504ns ± 3%  -11.96%  (p=0.032 n=5+5)
MulMod/mod192/uint256-6     195ns ± 4%     194ns ± 1%     ~     (p=0.841 n=5+5)
MulMod/mod192/big-6         563ns ± 6%     474ns ± 2%  -15.91%  (p=0.008 n=5+5)
MulMod/mod256/uint256-6     226ns ± 7%     124ns ± 3%  -45.02%  (p=0.008 n=5+5)
MulMod/mod256/big-6         518ns ± 5%     451ns ± 5%  -12.94%  (p=0.008 n=5+5)
```
So that also looks better than @gballet 's numbers

holiman:
> with go1.13.8 on Ubuntu 20.04.3 LTS.

@daosvik you should definitely update that, it's very old, and your benchmarks may be totally off.

gballet:
On an AWS `x5.large` instance, I get:

```
name                     old time/op    new time/op    delta
MulMod/small/uint256-4     44.0ns ± 0%    44.1ns ± 0%   +0.35%  (p=0.008 n=5+5)
MulMod/small/big-4         74.0ns ± 0%    74.8ns ± 0%   +1.06%  (p=0.008 n=5+5)
MulMod/mod64/uint256-4      127ns ± 0%     127ns ± 0%   -0.21%  (p=0.024 n=5+5)
MulMod/mod64/big-4          243ns ± 0%     245ns ± 0%   +0.75%  (p=0.008 n=5+5)
MulMod/mod128/uint256-4     207ns ± 0%     208ns ± 0%   +0.45%  (p=0.016 n=4+5)
MulMod/mod128/big-4         516ns ± 0%     520ns ± 0%   +0.73%  (p=0.008 n=5+5)
MulMod/mod192/uint256-4     212ns ± 0%     212ns ± 0%     ~     (p=0.087 n=5+5)
MulMod/mod192/big-4         493ns ± 0%     497ns ± 0%   +0.82%  (p=0.008 n=5+5)
MulMod/mod256/uint256-4     212ns ± 0%     132ns ± 0%  -37.80%  (p=0.000 n=5+4)
MulMod/mod256/big-4         466ns ± 0%     470ns ± 0%   +0.79%  (p=0.008 n=5+5)
```

holiman:
@daosvik another thing -- you're adding/changing some benchmarks. It would help if you moved those changes so that your very first commit is the benchmark-change, and after that, the commits that modify the code. 

That way, it's possible to compare benchmarks for "master" (master + new/modified benches) against PR, by just jumping between commits. Which is pretty difficult right now, when the bench-changes are mixed with code changes.

daosvik:
> @daosvik another thing -- you're adding/changing some benchmarks. It would help if you moved those changes so that your very first commit is the benchmark-change, and after that, the commits that modify the code.
> 
> That way, it's possible to compare benchmarks for "master" (master + new/modified benches) against PR, by just jumping between commits. Which is pretty difficult right now, when the bench-changes are mixed with code changes.

Sorry about that - I _almost_ got it right. The first commit adds an unnecessary printout of cache stats, which you can delete or comment out to generate the baseline benchmark.

gballet:
Right, so I hadn't seen that the resource-heavy go language server was running in the background. Without it, I do get results that are closer to expectation:

```
name                     old time/op    new time/op    delta
MulMod/small/uint256-8     52.2ns ±24%    52.5ns ±43%     ~     (p=1.000 n=5+5)
MulMod/small/big-8          113ns ± 5%     113ns ± 8%     ~     (p=1.000 n=5+5)
MulMod/mod64/uint256-8      168ns ± 9%     172ns ±13%     ~     (p=0.841 n=5+5)
MulMod/mod64/big-8          335ns ± 7%     364ns ±23%     ~     (p=0.222 n=5+5)
MulMod/mod128/uint256-8     278ns ± 4%     275ns ±18%     ~     (p=0.310 n=5+5)
MulMod/mod128/big-8         722ns ± 7%     738ns ±11%     ~     (p=0.690 n=5+5)
MulMod/mod192/uint256-8     291ns ± 6%     310ns ± 4%     ~     (p=0.095 n=5+5)
MulMod/mod192/big-8         685ns ± 7%     659ns ± 4%     ~     (p=0.548 n=5+5)
MulMod/mod256/uint256-8     295ns ± 8%     192ns ± 2%  -35.01%  (p=0.008 n=5+5)
MulMod/mod256/big-8         644ns ±10%     657ns ±13%     ~     (p=0.548 n=5+5)
```

mod.go
```diff
@@ -0,0 +1,669 @@
...
+// Some utility functions
+
+func add256(x, y Int, c uint64) (Int, uint64) {
+	var z Int
+	z[0], c = bits.Add64(x[0], y[0], c)
+	z[1], c = bits.Add64(x[1], y[1], c)
+	z[2], c = bits.Add64(x[2], y[2], c)
+	z[3], c = bits.Add64(x[3], y[3], c)
+	return z, c
+}
```
holiman, line 57: We already have 
```
func (z *Int) Add(x, y *Int) *Int {
	var carry uint64
	z[0], carry = bits.Add64(x[0], y[0], 0)
	z[1], carry = bits.Add64(x[1], y[1], carry)
	z[2], carry = bits.Add64(x[2], y[2], carry)
	z[3], _ = bits.Add64(x[3], y[3], carry)
	return z
}
```
and
```
// AddOverflow sets z to the sum x+y, and returns z and whether overflow occurred
func (z *Int) AddOverflow(x, y *Int) (*Int, bool) {
	var carry uint64
	z[0], carry = bits.Add64(x[0], y[0], 0)
	z[1], carry = bits.Add64(x[1], y[1], carry)
	z[2], carry = bits.Add64(x[2], y[2], carry)
	z[3], carry = bits.Add64(x[3], y[3], carry)
	return z, carry != 0
}
```
Seems like we already have the same code repeated twice, I'd like it if we don't repeat it a third time. We should be able to combine them, somehow

mod.go
```diff
@@ -0,0 +1,669 @@
...
+type cacheSet struct {
+	hit	uint64
+	miss	uint64
+
+	rw	sync.RWMutex
+	mod	[cacheWays]Int
+	inv	[cacheWays][5]uint64
+}
+
+var cache [cacheSets]cacheSet
```
holiman, line 38: I was wondering why the cache does not show up as allocs on the benchmark stats, but now I see: they're statically allocated. So the whole cache is `512` elements of `256` bytes each, and takes up a fixed chunk of 128K memory.

mod.go
```diff
@@ -0,0 +1,669 @@
...
+
+// reciprocal computes a 320-bit value representing 1/m
+//
+// Notes:
+// - reciprocal(0) = 0
+// - reciprocal(1) = 2^320-1
+// - otherwise, the result is normalised to have non-zero most significant word
+// - starts with a 32-bit division, refines with newton-raphson iterations
+
+func reciprocal(m Int) (mu [5]uint64) {
```
holiman, line 181: As this implementation is .. fairly nontrivial ... it would be good if you can add some links to the origin of the algorithm, so the astute reader can have something to verify against

mod.go
```diff
@@ -0,0 +1,669 @@
...
+			mu[i] = cache[cacheIndex].inv[w][i]
+		}
+
+		cache[cacheIndex].hit++
+		cache[cacheIndex].rw.RUnlock()
+
+		return
+	}
+
+	cache[cacheIndex].miss++
```
holiman, line 222: Isn't this an unsafe concurrent modification, since you're only holding the readlock ? `x++` is not atomic


(and if so, same a few lines above with `hit++`)

daosvik: Yes, it is unsafe, except it is not very important to have an exact result for these cache statistics.

holiman:
Another benchmark -- if the cache is disabled: 
```
name                     old time/op    new time/op    delta
MulMod/small/uint256-6     36.8ns ± 3%    36.6ns ± 3%     ~     (p=0.310 n=5+5)
MulMod/small/big-6         70.1ns ±13%    65.7ns ± 1%     ~     (p=0.310 n=5+5)
MulMod/mod64/uint256-6      117ns ± 9%     106ns ± 2%  -10.12%  (p=0.008 n=5+5)
MulMod/mod64/big-6          238ns ± 1%     220ns ± 5%   -7.70%  (p=0.008 n=5+5)
MulMod/mod128/uint256-6     211ns ± 9%     171ns ± 0%  -18.93%  (p=0.016 n=5+4)
MulMod/mod128/big-6         572ns ±11%     504ns ± 2%  -11.85%  (p=0.016 n=5+5)
MulMod/mod192/uint256-6     195ns ± 4%     196ns ± 2%     ~     (p=0.548 n=5+5)
MulMod/mod192/big-6         563ns ± 6%     483ns ± 5%  -14.20%  (p=0.008 n=5+5)
MulMod/mod256/uint256-6     226ns ± 7%     182ns ± 2%  -19.40%  (p=0.008 n=5+5)
MulMod/mod256/big-6         518ns ± 5%     451ns ± 3%  -12.81%  (p=0.008 n=5+5)
```
And the difference between with-cache and without-cache: 
```
[user@work uint256]$ benchstat daosvik.txt daosvik_nocache.txt 
name                     old time/op    new time/op    delta
MulMod/small/uint256-6     36.3ns ± 1%    36.6ns ± 3%     ~     (p=0.286 n=5+5)
MulMod/small/big-6         65.9ns ± 4%    65.7ns ± 1%     ~     (p=1.000 n=5+5)
MulMod/mod64/uint256-6      105ns ± 3%     106ns ± 2%     ~     (p=0.452 n=5+5)
MulMod/mod64/big-6          216ns ± 1%     220ns ± 5%     ~     (p=0.167 n=5+5)
MulMod/mod128/uint256-6     173ns ± 0%     171ns ± 0%   -0.92%  (p=0.029 n=4+4)
MulMod/mod128/big-6         504ns ± 3%     504ns ± 2%     ~     (p=0.690 n=5+5)
MulMod/mod192/uint256-6     194ns ± 1%     196ns ± 2%     ~     (p=0.548 n=5+5)
MulMod/mod192/big-6         474ns ± 2%     483ns ± 5%     ~     (p=0.310 n=5+5)
MulMod/mod256/uint256-6     124ns ± 3%     182ns ± 2%  +46.58%  (p=0.008 n=5+5)
MulMod/mod256/big-6         451ns ± 5%     451ns ± 3%     ~     (p=1.000 n=5+5)
```

mod.go
```diff
@@ -0,0 +1,669 @@
...
+			mu[0] = -m[0]
+		} else {
+			mu[4] = 0x8000000000000000 >> ((254-s) & 63)
+		}
+		return
+	}
+
+	// Check for reciprocal in the cache
+
+	cacheIndex := (m[3]^m[2]^m[1]^m[0]) & cacheMask
```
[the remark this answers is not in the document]
holiman: Then you could also move `cacheStats` and make it `cache.stats()`, which is nicer

mod.go
daosvik: You are right, and I have just removed the new ones.

mod.go
```diff
@@ -0,0 +1,669 @@
...
+		if cache[cacheIndex].mod[w][0] != m[0] { continue }
+
+		for i:=0; i<5; i++ {
+			mu[i] = cache[cacheIndex].inv[w][i]
+		}
+
+		cache[cacheIndex].hit++
+		cache[cacheIndex].rw.RUnlock()
+
+		return
```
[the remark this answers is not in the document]
daosvik: ok

mod.go
daosvik: Yes, that is correct.

mod.go
```diff
@@ -0,0 +1,669 @@
...
+//func shiftRight(x Int, s int) (z Int) {
+//	if s < 0 {
+//		z = shiftleft(x, uint(-s))
+//	} else {
+//		z = shiftright(x, uint(s))
+//	}
+//	return
+//}
+
+func shiftleft(x Int, s uint) (z Int) {
```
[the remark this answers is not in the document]
daosvik: We don't need both, but there was a performance improvement over Lsh on Ryzen, so I want to do some more benchmarking before removing one in favour of the other.

mod.go
daosvik: Ack. Third variant of addition (and subtraction) deleted.

holiman:
@chfast would you mind taking a peek at this too?

uint256.go
```diff
@@ -198,9 +198,29 @@ func (z *Int) AddMod(x, y, m *Int) *Int {
...
+	if z == m { // z is an alias for m and will be overwritten by AddOverflow before m is read
 		m = m.Clone()
 	}
+
+	// Fast path for m >= 2^192, with x and y at most slightly bigger than m.
+	// This is always the case when x and y are already reduced modulo such m.
+
+	if (m[3] != 0) && (m[3] >= x[3]) && (m[3] >= y[3]) {
+		if z, overflow := z.AddOverflow(x, y); overflow {
+			z.Sub(z, m)
```
chfast, line 210: Is this equivalent of https://github.com/chfast/intx/pull/206/files#diff-56937962f5980bbff23d21ab8da34df2b7df07b21546691156a2e7a79d763f15R1111-R1113?

uint256.go
```diff
@@ -198,9 +198,29 @@ func (z *Int) AddMod(x, y, m *Int) *Int {
...
+		}
+
+		for {
+			t := *z
+			if _, overflow := t.SubOverflow(&t, m); overflow {
+				break
```
chfast, line 216: How many times this loop can run?

uint256.go
```diff
@@ -198,9 +198,29 @@ func (z *Int) AddMod(x, y, m *Int) *Int {
...
 	}
-	if z == m { // z is an alias for m  // TODO: Understand why needed and add tests for all "division" methods.
+	if z == m { // z is an alias for m and will be overwritten by AddOverflow before m is read
 		m = m.Clone()
 	}
+
+	// Fast path for m >= 2^192, with x and y at most slightly bigger than m.
+	// This is always the case when x and y are already reduced modulo such m.
+
+	if (m[3] != 0) && (m[3] >= x[3]) && (m[3] >= y[3]) {
```
chfast, line 208: This only handles happy case, but maybe if `m[3] < x[3]` maybe it would be better to perform `x mod m` and then also use the more efficient implementation. Same for `y mod m`.

chfast, line 208:
```suggestion
	if (m[3] != 0) && (x[3] <= m[3]) && (y[3] <= m[3]) {
```

chfast (commented):
I only reviewed AddMod. I prefer to handle AddMod first as it is much simpler. Also benchmark changes can be merged first before functional changes.

I tried to optimize AddMod for similar cases but it made the worst case up to 2x slower. However, here the change is different so I will revisit this.

I also don't understand the motivation fully. Are ADDMOD and MULMOD so slow for the verkle-tree loads so slow that the node is not able to process blocks in time? If not, is gas cost reduction also planned?

uint256.go
daosvik: Not an exact equivalent, no. This code handles addition for m >= 2^192, with x,y <= m+2^192-1. This includes already-reduced values for x and y, that is 0 <= x,y < m.

uint256.go
daosvik: The worst case here is m = 2^192, x = y = 2m-1, so x+y = 4m-2. Hence the loop is exited at the latest in the fourth iteration. For already-reduced inputs, this happens no later than the second iteration.

uint256.go
daosvik: Happy case indeed, and also the normal case for modular arithmetic. :)

The code falls back to the slower division-based code for other cases, and will always return a fully-reduced result.

daosvik: If you and others feel that this is more readable, then it's ok with me.

chfast: Ok, I can see now we cannot efficiently use SubOverflow to reduce `x` here.

uint256.go
chfast: This comment should be added to the code.

Does this procedure has any name?

uint256.go
chfast: For me it is because I think of `x < m` and `y < m` as preconditions to efficient ADDMOD implementation.

holiman:
> Also benchmark changes can be merged first before functional changes.

Yeah, I noted the same. @daosvik can you make a separate PR with the benchmark changes, and then we can rebase this one on top of that one (or on top of master after we merge it). If not, I can do it myself too.

uint256.go
daosvik: Comment added.
This procedure doesn't have any particular name, it's just repeated subtraction of the modulus until we have the smallest non-negative residue.

gballet:
> I also don't understand the motivation fully. Are ADDMOD and MULMOD so slow for the verkle-tree loads so slow that the node is not able to process blocks in time? If not, is gas cost reduction also planned?

So this PR seeks to accelerate the specific case of doing modulo operations for a specific case, namely prime field operations used for elliptic curves. This isn't connected to the speed `ADDMOD` and `MULMOD` EVM instructions, but the calculations of the root commitment (equivalent to the root hash), so it can not really be addressed with gas costs.

mod.go
daosvik: I created this algorithm for this specific purpose, so there is no paper or website to refer to as the origin of the algorithm.
It's really only computing the reciprocal of the modulus with 320-bit precision, taking care to avoid overflow.

chfast:
> > I also don't understand the motivation fully. Are ADDMOD and MULMOD so slow for the verkle-tree loads so slow that the node is not able to process blocks in time? If not, is gas cost reduction also planned?
> 
> So this PR seeks to accelerate the specific case of doing modulo operations for a specific case, namely prime field operations used for elliptic curves. This isn't connected to the speed `ADDMOD` and `MULMOD` EVM instructions, but the calculations of the root commitment (equivalent to the root hash), so it can not really be addressed with gas costs.

To confirm: you want to use these methods outside of EVM.

chfast (changes requested):
- There is a bug in `AddMod()`. You can find failing tests in #89.
- The test `TestRandomMulMod` is running forever what indicates there may be also bug in `MulMod()`.
- @holiman, the CI testing does not run for external PR, this should be enabled.

holiman:
> @holiman, the CI testing does not run for external PR, this should be enabled.

I _think_ this is addressed now.

holiman:
Some remaining tasks for this PR:

- Rebase on top of latest master (benchmarks are now in `master`) 
- Move the caching implementation so it's not part of `reciprocal`
- Look into the test failures which @chfast reported

daosvik:
>     * There is a bug in `AddMod()`. You can find failing tests in [tests: more AddMod() test cases #89](https://github.com/holiman/uint256/pull/89).
> 
>     * The test `TestRandomMulMod` is running forever what indicates there may be also bug in `MulMod()`.
> 
>     * @holiman, the CI testing does not run for external PR, this should be enabled.



> Some remaining tasks for this PR:
> 
>     * Rebase on top of latest master (benchmarks are now in `master`)
> 
>     * Move the caching implementation so it's not part of `reciprocal`
> 
>     * Look into the test failures which @chfast reported

All of this is resolved with my latest force-push.
I left out the patch reworking the cache, as this can lead to cache thrashing by moduli mapping to the same way of a cache set

daosvik:
That should hopefully be my final force-push, making a clean baseline for benchmarks in #93 

Here are the key benchstat results comparing old and new operations using the new benchmarks:

Ryzen 5 3600, 3.6 GHz Zen2:
AddMod/mod256/uint256-12    34.2ns ± 0%    22.2ns ± 0%  -35.10%  (p=0.000 n=10+10)
MulMod/mod256/uint256-12     136ns ± 0%      72ns ± 1%  -47.04%  (p=0.000 n=9+10)

Raspberry Pi 4, 1.5 GHz Cortex-A72:
AddMod/mod256/uint256-4     147ns ± 0%      89ns ± 0%  -39.63%  (p=0.008 n=5+5)
MulMod/mod256/uint256-4     579ns ± 0%     459ns ± 0%  -20.74%  (p=0.008 n=5+5)

daosvik:
The latest commit (48c380f) is bad

mod.go
```diff
@@ -0,0 +1,574 @@
...
+)
+
+// Cache for reciprocal()
+//
+// Each cache set contains its own mutex to reduce lock contention in highly
+// multithreaded settings.
+//
+// Adjust cacheIndexBits and cacheWays to scale the size of the cache.
+// Reasonable values are quite small, e.g. cacheIndexBits from 2 to 10, and
+// cacheWays around 5 (5+8n makes set size an integer number of cachelines)
```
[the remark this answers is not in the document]
daosvik: Good idea. It's (24+72*cacheWays)*2^cacheIndexBits bytes, currently 96 KiB. Plus 16 bytes for hit&miss.

daosvik:
My latest force-push was rebased on the latest master with #93 , skipping the already-applied commits. It should all work now.

Regarding gofmt, please don't apply it to mod.go, as it will reduce readability and maintainability of the code. If you have specific requests about following Go style more closely, I will happily make the changes wherever it does not reduce readability significantly. It's great to have a tool help with maintaining at least a minimum standard of readability, but keeping this code readable requires more than gofmt provides.

holiman:
I added a fuzzer for this. cc @chfast -- if you're inclined to, you're very welcome to configure the mulmod fuzzer to run on CI (` go-fuzz -func Fuzz3Args`). Otherwise I can take a peek at it later. 

> Regarding gofmt, please don't apply it to mod.go, as it will reduce readability and maintainability of the code

I'm not sure about this. But if we do this, for selected methods, then they should be commented like this, so we can still run go format on the repo as a whole:

```
//nolint
func foo(){
...
}
```

chfast:
Ok, I will configure CI fuzzing later. On the subject I think we also need some fuzzing or at least a "random" unit test for AddMod.

chfast:
Some paths in mod.go are not covered by any test, even RandomMulMod (fuzzing is very ineffective to big int). You can see the coverage at https://app.codecov.io/gh/holiman/uint256/compare/86/tree/mod.go. Unit tests for these cases should be added.

chfast:
I ported AddMod implementation to intx (C++) and this works nicely although I will use branchless version even if less performant.

For the MulMod I will also try it some day, but I cannot say much about this implementation right now. But I recently started thinking about the "mu cache" embedded in the `MulMod` directly. If this is for specific use-case outside of EVM, can you consider different API design?

- `MulMod()` should not use cache, it may use reciprocal-base implementation if faster.
- `Reciprocal()` should be separate public function.
- `MulModWithReciprocal()` should take additional `mu` argument coming from `Reciprocal()`.

I'm assuming the verkle-tree use case always uses the same reciprocal so cache is not needed.

holiman:
I removed the fuzzer again, let's do that as a separate PR

holiman:
>  But if we do this, for selected methods, then they should be commented like this, so we can still run go format on the repo as a whole

Actually, that doesn't seem to work

holiman:
I'm leaning towards formatting everything. 
https://go.dev/blog/gofmt: 

> Gofmt’d code is:
> 
> -  easier to write: never worry about minor formatting concerns while hacking away,
> -   easier to read: when all code looks the same you need not mentally convert others' formatting style into something you can understand.
>-   easier to maintain: mechanical changes to the source don’t cause unrelated changes to the file’s formatting; diffs show only the real changes.
>-    uncontroversial: never have a debate about spacing or brace position ever again!

I think it's The Right Thing. If you think it makes certain things less readable, perhaps it can be rewritten using methods (which can be inlined later on) or scopes (just enclose snippets in brackets).

daosvik:
> I'm leaning towards formatting everything.
> https://go.dev/blog/gofmt:
> 
> > Gofmt’d code is:
> > 
> > * easier to write: never worry about minor formatting concerns while hacking away,
> > * easier to read: when all code looks the same you need not mentally convert others' formatting style into something you can understand.
> > * easier to maintain: mechanical changes to the source don’t cause unrelated changes to the file’s formatting; diffs show only the real changes.
> > * uncontroversial: never have a debate about spacing or brace position ever again!
> 
> I think it's The Right Thing. If you think it makes certain things less readable, perhaps it can be rewritten using methods (which can be inlined later on) or scopes (just enclose snippets in brackets).

I had read that article already, and respectfully both agree and disagree - gofmt'd code is indeed easier to write, read and maintain than a random collection of different style choices. However, for the particular case of mod.go, I would argue that _purely gofmt styled code_ is actually:
- harder to write
- harder to read, and
- harder to maintain.

Gofmt lengthens functions by insisting on only one statement per line, violating the principle of keeping functions reasonably short. This code also has examples where the standardised brace position of gofmt reduces readability. Additionally, according to that same article, 30% of "Go packages in the wild" are _not_ "formatted according to gofmt’s rules." So gofmt can be somewhat controversial too.

Unfortunately, using methods and/or scopes here is not likely to be very helpful, as each chunk of code is still going to lose its readability when gofmt'd.

Regarding The Right Thing, I'm inclined to agree with this quote from Google's C++ style guide: "As always, common sense and good taste should prevail." The guidelines for how to write code in a consistent style are very useful, but they are not strict rules that must be followed at all cost.

I also agree with the same style guide when it says to "Optimize for the reader, not the writer."

The gofmt'd version of mod.go is like looking through thick fog, and it would have been a nightmare to debug the code in this format.

When the original and gofmt'd versions of the code are placed side by side, it is much easier to see the structure in the original code, as the repetitions and differences from line to line stand out very clearly.

At the end of the day, I expect most people reading this code will do it to assure themselves that they can safely use it. It might be easier, as writers of the code, to just follow the Go tradition and gofmt the code. But in the long run, future readers will find it more difficult to read the code. My personal preference is to make the readers' task as easy as reasonably possible.

mod.go
```diff
@@ -0,0 +1,579 @@
...
+	return z
+}
+
+// reciprocal computes a 320-bit value representing 1/m
+//
+// Notes:
+// - reciprocal(0) = 0
+// - reciprocal(1) = 2^320-1
+// - otherwise, the result is normalised to have non-zero most significant word
+// - starts with a 32-bit division, refines with newton-raphson iterations
```
chfast, line 128: This computes `2^512/m` right? The Handbook of Applied Cryptography does not prescribe how to compute it. Why this methods has been chosen?

daosvik: > This computes `2^512/m` right? The Handbook of Applied Cryptography does not prescribe how to compute it. Why this methods has been chosen?

Yes. The basic idea is to start with a rough approximation and then repeatedly refine it with Newton-Raphson (see https://en.wikipedia.org/wiki/Division_algorithm#Newton%E2%80%93Raphson_division and http://www.quadibloc.com/comp/cp0202.htm). I looked at some other alternatives too before landing on this algorithm, since it is fast and does not involve any complicated bit manipulation. It is also used in hardware, starting with a small lookup table. The adaptations to this context are my own.

daosvik:
> I ported AddMod implementation to intx (C++) and this works nicely although I will use branchless version even if less performant.
> 
> For the MulMod I will also try it some day, but I cannot say much about this implementation right now. But I recently started thinking about the "mu cache" embedded in the `MulMod` directly. If this is for specific use-case outside of EVM, can you consider different API design?
> 
>     * `MulMod()` should not use cache, it may use reciprocal-base implementation if faster.
> 
>     * `Reciprocal()` should be separate public function.
> 
>     * `MulModWithReciprocal()` should take additional `mu` argument coming from `Reciprocal()`.
> 
> 
> I'm assuming the verkle-tree use case always uses the same reciprocal so cache is not needed.

A cache is indeed not always needed, but we should not expose too much of the inner workings of the implementation in its external interface. Please have a look at my latest commits, and tell me how you like this approach where users can configure the module with or without a fixed modulus+reciprocal and with or without a regular cache.

I also added tests to increase the code coverage, but I don't yet know how to trigger codecov, so maybe you could do that?

holiman:
> but I don't yet know how to trigger codecov, so maybe you could do that?

codecode triggers automatically, however, it updates it's previous comment instead of writing a new one. Link for this PR: https://app.codecov.io/gh/holiman/uint256/compare/86/diff

mod.go
```diff
@@ -0,0 +1,624 @@
...
+// Adjust cacheIndexBits and cacheWays to scale the size of the cache.
+// Total cache size is (24+72*cacheWays)*2^cacheIndexBits bytes, which is
+// 96 KiB with cacheIndexBits = 8 and cacheWays = 5. There are also 16 bytes
+// for hit and miss counters.
+//
+// Reasonable values are quite small, e.g. cacheIndexBits from 2 to 10, and
+// cacheWays around 5. Note that cacheWays = 5+8n makes the set size an integer
+// number of 64-byte cachelines.
+//
+// If you want to disable the cache, set cacheIndexBits and cacheWays to 0
```
holiman, line 26: This doesn't really work, at least not if the 'consumer' of the library wants to use it as a go module. The only way to make this work is to copy-paste the entire library and modify the value, since it's a `const`.

holiman:
Regarding this:

Screenshot 2021-09-06 at 08-53-13 Codecov

I put a `panic` there, and modified the fuzzer to just generate a number and then call `reciprocal`, to find some value to trigger it and then add a testcase. Been fuzzing for five minutes, haven't found anything to trigger that path.

mod.go
```diff
@@ -0,0 +1,624 @@
...
+
+	// 0 or a power of 2?
+
+	if onesCount(m) <= 1 {
+		if s >= 255 { // m <= 1
+			mu[4] = -m[0]
+			mu[3] = -m[0]
+			mu[2] = -m[0]
+			mu[1] = -m[0]
+			mu[0] = -m[0]
```
holiman, line 188: Unreachable via mulmod, which only calls `reciprocal` if `m[3]!=0`, so `leadingZeroes(m)` can never be `>63`. 

Either we need specific tests on `reciprocal` to cover this, or this can be removed, at a loss of generality.

holiman: I tried to add this: 
```golang
	t.Run("CustomMulMod", func(t *testing.T) {
		// CustomMulMod is a special case. Normally, we don't use the
		// reciprocal-approach if the modulus is < 192 bits. This
		// uses the reciprocal-approach in all cases, for better
		// coverage

		mulmodCustom := func(z, x, y, m *Int) *Int {
			if x.IsZero() || y.IsZero() || m.IsZero() {
				return z.Clear()
			}
			p := umul(x, y)
			mu := reciprocal(*m)
			r := reduce4(p, *m, mu)
			return z.Set(&r)
		}
		proc(t, mulmodCustom, func(z, x, y, m *big.Int) *big.Int {
			if m.Sign() == 0 {
				return z.SetUint64(0)
			}
			return mulMod(z, x, y, m)
		})
	})
```
But that causes the test to hang. I'm not sufficiently good at math to figure out how to do it correctly

mod.go
daosvik: The constness allows the compiler to allocate the cache memory statically and to remove unused code paths. Does Go have a construct that can replace templates, allowing users to instantiate the cache with their own parameters? Or maybe the cache needs to be dynamically allocated to give users this flexibility, at the risk of losing performance?

holiman:
Yes and no. It can probably be worked around by hacking the compiler directives, but I think it's not go-idiomatic to do so. Then dependent projects would need to add the specific build-tag to their pipeline, which is pretty messy. I think it's better to use a dynamically allocated cache. We could provide the primitives for that without exposing the internals, just the public methods (which, for cache, can be none at all). Then we could either have it be a "user choose size" or just a boolean, cache or no cache. Performance-wise, I'm not sure it has to have a significant impact. If the cache type is internal, and we use non-public methods, then the compiler should still be able to inline the calls.

holiman:
I pushed an alternative implementation of the cache, making it so the caller first creates a cache (which can be a global var on the caller side), and then supplies it in the call to mulmod. 
For the big mulmods, I see no regression, but it does add a little overhead to the smaller mulmods (for some reason) 
```
MulMod/small/uint256-6   33.4ns ± 4%  41.2ns ± 8%  +23.52%  (p=0.008 n=5+5)
MulMod/mod64/uint256-6   75.9ns ± 3%  96.6ns ± 4%  +27.29%  (p=0.008 n=5+5)
MulMod/mod128/uint256-6   112ns ± 1%   140ns ±12%  +25.01%  (p=0.008 n=5+5)
MulMod/mod192/uint256-6   167ns ±22%   156ns ± 1%     ~     (p=0.730 n=5+4)
MulMod/mod256/uint256-6  98.3ns ±13%  97.3ns ±10%     ~     (p=0.548 n=5+5)
```
Those tests are _with_ cache

holiman:
Regarding the formatting. We need to format this, for maintenance reasons. Otherwise it'll just happen by accident some time in the future, when some other committer wants to modify some detail somewhere. Since it's not possibly to simply denote a block of code as "don't format", we'll just have to do the entire file. Sorry :(

holiman:
Moved the cache to a separate file, and un-exported the type. The latter change improved the figures a bit: 
```
MulMod/small/uint256-6   33.4ns ± 4%  37.5ns ± 8%  +12.46%  (p=0.008 n=5+5)
MulMod/mod64/uint256-6   75.9ns ± 3%  85.2ns ± 5%  +12.28%  (p=0.008 n=5+5)
MulMod/mod128/uint256-6   112ns ± 1%   127ns ± 5%  +13.48%  (p=0.008 n=5+5)
MulMod/mod192/uint256-6   167ns ±22%   154ns ± 0%     ~     (p=0.690 n=5+5)
MulMod/mod256/uint256-6  98.3ns ±13%  93.8ns ±11%     ~     (p=0.095 n=5+5)
```
(global cache versus caller-provided cache)

reciprocal_cache.go
```diff
@@ -0,0 +1,134 @@
...
+func init() {
+	if fixedModulus {
+		// Initialise fixed modulus
+		fixed_m[3] = 0xffffffff00000001
+		fixed_m[2] = 0x0000000000000000
+		fixed_m[1] = 0x00000000ffffffff
+		fixed_m[0] = 0xffffffffffffffff
+
+		fixed_r = reciprocal(fixed_m, nil)
+	}
```
holiman, line 77: This is a bit magic to me. What is that fixed modulus? Please add a comment about what it represents. 
Also, this is another case of where, if someone wants to use a different fixed modulus, they have to copy-paste the entire library and cannot use it as a go module. 

I think it may be better to define certain fixed modulus's, and hten let the caller either use one of those or supply one himself. 
Like
```
var (
	// Fixed modulus and its reciprocal
	FixedModulusCurveFoo , _ = FromHex("0xffffffff00000001000000000000000000000000ffffffffffffffffffffffff")
	FixedReciprocalCurveFoo [5]uint64
)

func init() {
	FixedReciprocalCurveFoo = reciprocal(*FixedModulusCurveFoo, nil)
}
```
And then the cache itself could be initalized with a the fixed modulus to use, if any

holiman: Pushed a commit to address this, consider it a suggestion.

holiman:
Even without the cache, I still see a speed up: 
```
MulMod/small/uint256-6   32.0ns ± 1%  32.4ns ± 4%     ~     (p=0.064 n=8+10)
MulMod/mod64/uint256-6   74.3ns ± 1%  75.7ns ± 1%   +1.87%  (p=0.000 n=8+9)
MulMod/mod128/uint256-6   122ns ± 2%   121ns ± 3%     ~     (p=0.092 n=9+8)
MulMod/mod192/uint256-6   149ns ± 4%   148ns ± 0%     ~     (p=0.813 n=9+9)
MulMod/mod256/uint256-6   187ns ± 5%   160ns ± 9%  -14.26%  (p=0.000 n=9+10)
MulMod/small/big-6       61.3ns ±17%  57.2ns ± 1%     ~     (p=0.780 n=10+9)
MulMod/mod64/big-6        112ns ± 9%   113ns ± 6%     ~     (p=0.725 n=10+10)
MulMod/mod128/big-6       365ns ± 7%   357ns ± 2%     ~     (p=0.182 n=10+9)
MulMod/mod192/big-6       408ns ± 1%   416ns ± 6%     ~     (p=0.779 n=9+9)
MulMod/mod256/big-6       473ns ± 4%   485ns ± 6%     ~     (p=0.280 n=10+10)
```

holiman:
And also for addmod: 
```
[user@work uint256]$ benchstat addmod_before.txt addmod_after.txt 
AddMod/small/uint256-6   12.0ns ± 2%  12.2ns ± 2%   +1.65%  (p=0.016 n=10+9)
AddMod/mod64/uint256-6   19.3ns ± 3%  19.4ns ± 0%     ~     (p=0.055 n=9+8)
AddMod/mod128/uint256-6  43.3ns ± 2%  42.0ns ± 6%     ~     (p=0.661 n=9+10)
AddMod/mod192/uint256-6  46.6ns ± 1%  46.9ns ± 1%   +0.59%  (p=0.023 n=9+9)
AddMod/mod256/uint256-6  49.8ns ± 1%  19.8ns ± 4%  -60.19%  (p=0.000 n=9+10)
AddMod/small/big-6       44.7ns ± 1%  43.4ns ± 1%   -2.92%  (p=0.000 n=9+9)
AddMod/mod64/big-6       54.8ns ± 1%  53.6ns ± 1%   -2.14%  (p=0.000 n=9+9)
AddMod/mod128/big-6       143ns ± 3%   141ns ± 1%   -1.87%  (p=0.001 n=10+9)
AddMod/mod192/big-6       144ns ± 2%   143ns ± 2%     ~     (p=0.305 n=10+9)
AddMod/mod256/big-6       144ns ± 1%   143ns ± 1%     ~     (p=0.868 n=9+8)
```

holiman (approved):
I think this is OK now. Some follow-up tasks:
 
- Adding the `MulmodWithReciprocal` and `Reciprocal` methods.
- We should integrate `uint256` with oss-fuzz, and let it run for a while before cutting a new release. 
- `go fmt` eventually

uint256.go
```diff
@@ -576,6 +609,38 @@ func (z *Int) SMod(x, y *Int) *Int {
...
+		return z.Clear()
+	}
+	p := umul(x, y)
+
+	if m[3] != 0 {
+		r := reduce4(p, m, *mu)
+		return z.Set(&r)
+	}
+
+	var (
```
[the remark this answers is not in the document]
daosvik: It has been modified so that it is limited to m[3] != 0. Hence the first comment inside the function is "Note: specialized for m[3] != 0".
It's not so clear this way, so I'll adjust the comments preceding the function.

PR #88, opened 2021-08-23
benchmarks: use least non negative residue of the addend for each modulus when testing AddMod

daosvik:
Separate PR for benchmark changes separated out from #86 as requested


holiman (approved):
LGTM

PR #95, opened 2021-08-31
fuzz: add fuzzer for mulmod

holiman:
Fuzzer for mulmod

### DIFF
--- a/fuzz.go
+++ b/fuzz.go
@@ -166,26 +166,26 @@ func checkThreeArgOp(op opThreeArgFunc, bigOp bigThreeArgFunc, x, y, z Int) {
 }
 
 func Fuzz(data []byte) int {
-	if len(data) != 65 {
-		return -1
+	switch len(data) {
+	case 64:
+		return fuzzBinaryOp(data)
+	case 96:
+		return fuzzTernaryOp(data)
 	}
-	op := data[0]
+	return -1
+}
+func fuzzBinaryOp(data []byte) int {
 	var x, y Int
-	x.SetBytes(data[1:33])
-	y.SetBytes(data[33:])
-
-	switch op {
-	case opUdivrem:
-		if y.IsZero() {
-			return 0
-		}
+	x.SetBytes(data[0:32])
+	y.SetBytes(data[32:])
+	if !y.IsZero() { // uDivrem
 		checkDualArgOp((*Int).Div, (*big.Int).Div, x, y)
 		checkDualArgOp((*Int).Mod, (*big.Int).Mod, x, y)
-
-	case opMul:
+	}
+	{ // opMul
 		checkDualArgOp((*Int).Mul, (*big.Int).Mul, x, y)
-
-	case opLsh:
+	}
+	{ // opLsh
 		lsh := func(z, x, y *Int) *Int {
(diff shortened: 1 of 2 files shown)


fuzz.go
```diff
@@ -120,14 +196,46 @@ func Fuzz(data []byte) int {
...
+func bigMulMod(b1, b2, b3, b4 *big.Int) *big.Int {
+	return b1.Mod(big.NewInt(0).Mul(b2, b3), b4)
+}
+
+func intMulMod(f1, f2, f3, f4 *Int) *Int {
+	return f1.MulMod(f2, f3, f4)
+}
+
+func Fuzz3Args(data []byte) int {
+	if len(data) != 1+32+32+23 {
```
chfast, line 221:
```suggestion
	if len(data) != 1+32+32+32 {
```

fuzz.go
```diff
@@ -120,14 +196,46 @@ func Fuzz(data []byte) int {
...
 	}
+	op := data[0]
 
-	return 0
+	var x, y, z Int
+	x.SetBytes(data[1:33])
+	y.SetBytes(data[33:65])
+	z.SetBytes(data[65:])
+
+	switch op {
```
chfast, line 231: I was actually thinking that maybe instead of switching on the first byte just run all operators in a row.

fuzz.go
```diff
@@ -120,14 +196,46 @@ func Fuzz(data []byte) int {
...
+	case opMulmod:
+		if z.IsZero() {
+			return 0
+		}
+		checkThreeArgOp(intMulMod, bigMulMod, x, y, z)
```
chfast, line 236: Add also `AddMod`.

fuzz.go
```diff
@@ -2,6 +2,7 @@
 // Copyright 2020 uint256 Authors
 // SPDX-License-Identifier: BSD-3-Clause
 
+//go:build gofuzz
```
chfast, line 5: What is this?

fuzz.go
```diff
@@ -120,14 +196,46 @@ func Fuzz(data []byte) int {
...
+
+func bigMulMod(b1, b2, b3, b4 *big.Int) *big.Int {
+	return b1.Mod(big.NewInt(0).Mul(b2, b3), b4)
+}
+
+func intMulMod(f1, f2, f3, f4 *Int) *Int {
+	return f1.MulMod(f2, f3, f4)
+}
+
+func Fuzz3Args(data []byte) int {
```
chfast, line 220: Having separate function confuses go-fuzz. This is not needed, you can switch between binary and ternary operators by the length of the input data.

fuzz.go
holiman: Yes, good idea

fuzz.go
holiman: go 1.17 build directives have changed. go formt auto-converts the old style to the new one

fuzz.go
holiman: Sure, we could do that instead

holiman:
Fixed. I guess the old corpus is moot now though, so we could clear that from CI (dunno how much it matters)

chfast (approved):

PR #110, opened 2022-02-21
Add Muldiv()

Planxnx:
`(x * y) / denominator` is the most common case that cannot be calculated with full precision by normal methods `z.Mul(x, y).Div(z, denominator)` 

ex. `(max_uint256 * 2)/10` 
an overflow occurs when you try to calculated by normal methods, but the result is smaller than `max_uint256`.

`MulDivOverflow()` is 512-bit multiplication and 512-bit by 256-bit division.
returns The 256-bit result and overflow occurred. 

inspired and realworld use-case from UniswapV3
[https://github.com/Uniswap/v3-core/blob/main/contracts/libraries/FullMath.sol](url)

```
BenchmarkMulDivOverflow/small/uint256-8         	491599033	         2.297 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div64/uint256-8         	512714014	         2.351 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div128/uint256-8        	488145638	         2.415 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div192/uint256-8        	484725356	         2.457 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div256/uint256-8        	438809512	         2.509 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/small/big-8             	58039941	        20.62 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div64/big-8             	58061472	        20.61 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div128/big-8            	43477892	        24.84 ns/op	       1 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div192/big-8            	46356553	        24.83 ns/op	       2 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div256/big-8            	40076848	        28.15 ns/op	       4 B/op	       0 allocs/op
```

### DIFF
--- a/fuzz.go
+++ b/fuzz.go
@@ -215,7 +215,7 @@ func Fuzz(data []byte) int {
 		return fuzzTernaryOp(data) // needs 96 byte
 	}
 	// Too large input
-	return 0
+	return -1
 }
 
 func fuzzUnaryOp(data []byte) int {
--- a/fuzz.go
+++ b/fuzz.go
@@ -24,6 +24,9 @@ const (
 	opMulmod
 )
 
+type opUnaryArgFunc func(*Int, *Int) *Int
+type bigUnaryArgFunc func(*big.Int, *big.Int) *big.Int
+
 type opDualArgFunc func(*Int, *Int, *Int) *Int
 type bigDualArgFunc func(*big.Int, *big.Int, *big.Int) *big.Int
 
@@ -40,6 +43,40 @@ func crash(op interface{}, msg string, args ...Int) {
 	}
 	panic(fmt.Sprintf("%s\nfor %s (%s:%d)\n%v",
 		msg, fnName, fnFile, fnLine, strings.Join(strArgs, "\n")))
+}
+
+func checkUnaryOp(op opUnaryArgFunc, bigOp bigUnaryArgFunc, x Int) {
+	origX := x
+	var result Int
+	ret := op(&result, &x)
+	if ret != &result {
+		crash(op, "returned not the pointer receiver", x)
+	}
+	if x != origX {
+		crash(op, "argument modified", x)
+	}
+	expected, _ := FromBig(bigOp(new(big.Int), x.ToBig()))
+	if result != *expected {
+		crash(op, "unexpected result", x)
+	}
+	// Test again when the receiver is not zero.
+	var garbage Int
+	garbage.Sub(&garbage, NewInt(1))
+	ret = op(&garbage, &x)
+	if ret != &garbage {
+		crash(op, "returned not the pointer receiver", x)
+	}
+	if garbage != *expected {
+		crash(op, "unexpected result", x)
+	}
+	// Test again with the receiver aliasing arguments.
+	ret = op(&x, &x)
+	if ret != &x {
+		crash(op, "returned not the pointer receiver", x)
+	}
+	if x != *expected {
+		crash(op, "unexpected result", x)
+	}
 }
 
 func checkDualArgOp(op opDualArgFunc, bigOp bigDualArgFunc, x, y Int) {
@@ -166,7 +203,16 @@ func checkThreeArgOp(op opThreeArgFunc, bigOp bigThreeArgFunc, x, y, z Int) {
 }
 
 func Fuzz(data []byte) int {
+	if len(data) < 32 {
+		return 0
+	} else {
+
+		return fuzzUnaryOp(data)
+	}
+
 	switch len(data) {
+	case 32:
+		return fuzzUnaryOp(data)
 	case 64:
 		return fuzzBinaryOp(data)
 	case 96:
@@ -174,6 +220,14 @@ func Fuzz(data []byte) int {
 	}
 	return -1
 }
+
+func fuzzUnaryOp(data []byte) int {
+	var x Int
+	x.SetBytes(data[0:32])
+	checkUnaryOp((*Int).Sqrt, (*big.Int).Sqrt, x)
+	return 1
+}
+
 func fuzzBinaryOp(data []byte) int {
 	var x, y Int
 	x.SetBytes(data[0:32])
--- a/fuzz.go
+++ b/fuzz.go
@@ -299,7 +299,7 @@ func fuzzTernaryOp(data []byte) int {
 	{ // addMod
 		checkThreeArgOp(intAddMod, bigAddMod, x, y, z)
 	}
-	{ // addMod
+	{ // mulDiv
 		checkThreeArgOp(intMulDiv, bigMulDiv, x, y, z)
 	}
 	return 1
--- a/fuzz.go
+++ b/fuzz.go
@@ -205,20 +205,17 @@ func checkThreeArgOp(op opThreeArgFunc, bigOp bigThreeArgFunc, x, y, z Int) {
 func Fuzz(data []byte) int {
 	if len(data) < 32 {
 		return 0
-	} else {
-
-		return fuzzUnaryOp(data)
 	}
-
-	switch len(data) {
-	case 32:
-		return fuzzUnaryOp(data)
-	case 64:
-		return fuzzBinaryOp(data)
-	case 96:
-		return fuzzTernaryOp(data)
+	switch {
+	case len(data) < 64:
+		return fuzzUnaryOp(data) // needs 32 byte
+	case len(data) < 96:
+		return fuzzBinaryOp(data) // needs 64 byte
+	case len(data) < 128:
+		return fuzzTernaryOp(data) // needs 96 byte
 	}
-	return -1
+	// Too large input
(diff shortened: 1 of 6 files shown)


holiman:
I pushed a fuzzer to this PR, looks good so far. Also, I noticed that the fuzzer was broken, which is also fixed.

fuzz.go
```diff
@@ -292,5 +299,8 @@ func fuzzTernaryOp(data []byte) int {
 	{ // addMod
 		checkThreeArgOp(intAddMod, bigAddMod, x, y, z)
 	}
+	{ // addMod
```
holiman, line 302:
```suggestion
	{ // mulDiv
```

Planxnx:
I've optimized it a bit by zero-check on the p[4:] directly instead copy to new Int.
```
BenchmarkMulDivOverflow/small/uint256-8         	474043095	         2.285 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div64/uint256-8         	501923428	         2.294 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div128/uint256-8        	498558622	         2.374 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div192/uint256-8        	484291431	         2.404 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div256/uint256-8        	441282002	         2.491 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/small/big-8             	58109386	        20.55 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div64/big-8             	58929327	        20.56 ns/op	       0 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div128/big-8            	43171965	        24.07 ns/op	       1 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div192/big-8            	45164564	        24.88 ns/op	       2 B/op	       0 allocs/op
BenchmarkMulDivOverflow/div256/big-8            	41473440	        26.91 ns/op	       4 B/op	       0 allocs/op
```

holiman:
cc @chfast would you mind taking a look?

Planxnx:
@holiman Is everything okay? How long will it take?

uint256.go
```diff
@@ -673,6 +673,29 @@ func (z *Int) MulMod(x, y, m *Int) *Int {
 	return z.Set(&rem)
 }
 
+// MulDivOverflow calculates (x*y)/d with full precision, returns z and whether overflow occurred.
```
chfast, line 676: Please explain what "overflow" means in this case: the division result does not fit 256-bit return type.

uint256.go
```diff
@@ -673,6 +673,29 @@ func (z *Int) MulMod(x, y, m *Int) *Int {
...
+// computes 512-bit multiplication and 512 by 256 division.
+func (z *Int) MulDivOverflow(x, y, d *Int) (*Int, bool) {
+	if x.IsZero() || y.IsZero() || d.IsZero() {
+		return z.Clear(), false
+	}
+	p := umul(x, y)
+
+	// If the multiplication is within 256 bits use Div().
+	if (p[4] | p[5] | p[6] | p[7]) == 0 {
```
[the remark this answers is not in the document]
chfast: This requires a fix to `udivrem()` to handle case when `uLen < dLen` (quick return).
Benchmarks (current PR vs simplified version):
```
name                             old time/op  new time/op  delta
MulDivOverflow/small/uint256-8   2.23ns ± 0%  2.24ns ± 1%  +0.66%  (p=0.000 n=9+10)
MulDivOverflow/div64/uint256-8   2.25ns ± 0%  2.25ns ± 0%  -0.05%  (p=0.008 n=8+8)
MulDivOverflow/div128/uint256-8  2.37ns ± 0%  2.36ns ± 0%  -0.51%  (p=0.000 n=9+9)
MulDivOverflow/div192/uint256-8  2.41ns ± 0%  2.39ns ± 0%  -0.64%  (p=0.000 n=8+9)
MulDivOverflow/div256/uint256-8  2.56ns ± 0%  2.54ns ± 0%  -0.66%  (p=0.000 n=8+9)
MulDivOverflow/small/big-8       18.4ns ± 0%  19.0ns ± 0%  +3.16%  (p=0.000 n=9+10)
MulDivOverflow/div64/big-8       18.4ns ± 0%  19.0ns ± 1%  +3.22%  (p=0.000 n=9+10)
MulDivOverflow/div128/big-8      22.2ns ± 1%  22.9ns ± 1%  +3.10%  (p=0.000 n=10+10)
MulDivOverflow/div192/big-8      23.2ns ± 1%  23.8ns ± 1%  +2.78%  (p=0.000 n=9+9)
MulDivOverflow/div256/big-8      26.7ns ± 1%  27.4ns ± 1%  +2.69%  (p=0.000 n=10+10)
```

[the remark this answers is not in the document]
Planxnx: I've removed `Div()` and added a handle case where the denominator is greater than the numerator instead.

Planxnx:
@chfast I've updated the code. Can you do another review?

holiman:
The last change made the code coverage from the tests drop a bit - see https://app.codecov.io/gh/holiman/uint256/compare/110/diff . 

Screenshot 2022-03-08 at 20-05-39 Codecov

Not sure if I think the `pl.Eq(d)` case warrants a separate check, I would rather just remove it personally.

uint256.go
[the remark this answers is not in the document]
chfast: My idea is to handle this case inside `udivrem()`. My attempt is in https://github.com/holiman/uint256/pull/111 but we need to figure out why one of the tests fails (help welcome).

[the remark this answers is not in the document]
chfast: The #111 looks to be ready. Can you try to rebase on top of it and completely remove this block?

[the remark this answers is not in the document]
Planxnx: done ✅
@chfast  thanks !!

fuzz.go
```diff
@@ -205,20 +205,17 @@ func checkThreeArgOp(op opThreeArgFunc, bigOp bigThreeArgFunc, x, y, z Int) {
...
-	case 32:
-		return fuzzUnaryOp(data)
-	case 64:
-		return fuzzBinaryOp(data)
-	case 96:
-		return fuzzTernaryOp(data)
-	}
-	return -1
+	switch {
+	case len(data) < 64:
```
chfast, line 210: Why this change?

fuzz.go
```diff
@@ -277,6 +274,16 @@ func intAddMod(f1, f2, f3, f4 *Int) *Int {
 	return f1.AddMod(f2, f3, f4)
 }
 
+func bigMulDiv(b1, b2, b3, b4 *big.Int) *big.Int {
+	b1.Mul(b2, b3)
+	return b1.Div(b1, b4)
+}
+
+func intMulDiv(f1, f2, f3, f4 *Int) *Int {
```
chfast, line 282: This may be upgraded to public function in future so maybe change it to `func (z *Int) mulDiv(x, y, d *Int) *Int`?

fuzz.go
```diff
@@ -205,20 +205,17 @@ func checkThreeArgOp(op opThreeArgFunc, bigOp bigThreeArgFunc, x, y, z Int) {
...
+	switch {
+	case len(data) < 64:
+		return fuzzUnaryOp(data) // needs 32 byte
+	case len(data) < 96:
+		return fuzzBinaryOp(data) // needs 64 byte
+	case len(data) < 128:
+		return fuzzTernaryOp(data) // needs 96 byte
+	}
+	// Too large input
+	return 0
```
chfast, line 218: This should still return -1 I believe.

chfast (commented):
Looks good to me with some questions around testing.

fuzz.go
holiman: I figured it would be better to use the input we have, rather than be too picky about the input size...?

fuzz.go
holiman: Yes, you're right

fuzz.go
holiman: Let's do that later in that case

fuzz.go
chfast: I'm not sure this is good idea because you have unused bytes in the end of the input. They will give you exactly the same coverage path and also for all inputs giving the same coverage fuzzers prefer the shortest one and should reduce corpus entries finally. So the end result should be the same but this may be just longer route to get there.

I will experiment with two versions to see which one is faster.

PR #112, opened 2022-03-17
test: Refactor tests

chfast:
Rename big.Int reference operator implementations with prefix "big" and unify their usage in random tests.


holiman (approved):
LGTM

Issue #103, opened 2021-12-03
Include `DivMod()` method


2opremio:
We have started using https://github.com/holiman/uint256 at https://github.com/stellar/go and we are missing the [`DivMod()`](https://pkg.go.dev/math/big#Int.DivMod) method.

This has required invoking `Div()` and `Mod()` separately, which isn't great for performance.

PR #113, opened 2022-03-22
Add DivMod()

chfast:
DivMod() sets z to the quotient x div y and m to the modulus x mod y and returns the pair (z, m) for y != 0.
If y == 0, both z and m are set to 0 (OBS: differs from the big.Int).

Closes https://github.com/holiman/uint256/issues/103.


holiman (approved):
LGTM

PR #114, opened 2022-03-25
fuzz: Tune fuzzing input selection

chfast:
Running with default options, the last file added to corpus (then I waited additional 5 mins).
_before_:
```
2022/03/25 08:24:18 workers: 8, corpus: 126 (1s ago), crashers: 0, restarts: 1/9925, execs: 13409730 (74498/sec), cover: 578, uptime: 3m0s                              
```
_after_:
```
2022/03/25 08:31:49 workers: 8, corpus: 114 (1s ago), crashers: 0, restarts: 1/9771, execs: 5403616 (78313/sec), cover: 578, uptime: 1m9s                               
```

### DIFF
--- a/fuzz.go
+++ b/fuzz.go
@@ -203,18 +203,15 @@ func checkThreeArgOp(op opThreeArgFunc, bigOp bigThreeArgFunc, x, y, z Int) {
 }
 
 func Fuzz(data []byte) int {
-	if len(data) < 32 {
-		return 0
-	}
-	switch {
-	case len(data) < 64:
+	switch len(data) {
+	case 32:
 		return fuzzUnaryOp(data) // needs 32 byte
-	case len(data) < 96:
+	case 64:
 		return fuzzBinaryOp(data) // needs 64 byte
-	case len(data) < 128:
+	case 96:
 		return fuzzTernaryOp(data) // needs 96 byte
 	}
-	// Too large input
+	// Invalid input length
 	return -1
 }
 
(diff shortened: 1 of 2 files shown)


fuzz.go
```diff
@@ -203,18 +203,15 @@ func checkThreeArgOp(op opThreeArgFunc, bigOp bigThreeArgFunc, x, y, z Int) {
 }
 
 func Fuzz(data []byte) int {
-	if len(data) < 32 {
-		return 0
-	}
-	switch {
-	case len(data) < 64:
+	switch len(data) {
+	case 64:
```
holiman, line 207:
```suggestion
	case 32:
```

fuzz.go
```diff
@@ -203,18 +203,15 @@ func checkThreeArgOp(op opThreeArgFunc, bigOp bigThreeArgFunc, x, y, z Int) {
...
 		return fuzzUnaryOp(data) // needs 32 byte
-	case len(data) < 96:
+	case 96:
```
holiman, line 209:
```suggestion
	case 64:
```

fuzz.go
```diff
@@ -203,18 +203,15 @@ func checkThreeArgOp(op opThreeArgFunc, bigOp bigThreeArgFunc, x, y, z Int) {
...
 		return fuzzBinaryOp(data) // needs 64 byte
-	case len(data) < 128:
+	case 128:
```
holiman, line 211:
```suggestion
	case 96:
```

holiman (changes requested):

chfast:
This has some issues so I don't want to mess up with it.

Issue #108, opened 2022-01-31
SetHex or SetString equivalent


t-anyu:
Hey,

it would be nice to have a math/big.Int SetString equivalent so that we can re-use Int objects without releasing to GC. Maybe just expose  (z *Int) fromHex(hex string)?

Thanks

holiman:
Well, there's always 
```
// UnmarshalText implements encoding.TextUnmarshaler
func (z *Int) UnmarshalText(input []byte) error {
	return z.fromHex(string(input))
}

```
But perhaps you want the ability to do `SetString("500", 10)` and choose the base? 

I don't really se any problem with changing `fromHex` into `FromHex`, tbh. Although, I guess it cannot be exposed in it's current form -- as we have it now, the `FromHex` guarantees that the `z` is clear, and thus it doesn't have to set all components (only the ones that are used). If we expose it, we need to ensure to clear any prior bits.

PR #122, opened 2022-09-25
[wip] SetString (#108)

elee1766:
hi, i was in need of two things

1. scanner + valuer interfaces for postgresql Numeric
2. base 10 string -> uint256.Int 

i saw that issue #108 was a thing, and so I created a fromBase10 and incomplete SetString(s string, base int)  (err error), since it can't do binary (0b prefix) 

not really sure where to post this, but i thought a PR would be good at least so you could see the code changes. i don't really know what interface you are targeting, maybe setstring should return a bool to match big.Int? 

big.Int's setString has an alloc in it, so I thought i would get rid of it since i saw the library was promising 0 alloc. i did a super stupid and naive implementation for base 10 parsing, since it was something i needed in another project i was working on, and I just needed something that worked. i've been using this fork and it's been working fine (at least for now)

if you are curious, the benchmarks are here. so it's better than std but it's still not great... i'll be working on improving it, will probably require a rethink lol

base16 is the existing hex implementation.

```
go test -run=String -bench=String

goos: linux
goarch: amd64
cpu: Intel(R) Core(TM) i9-9900KF CPU @ 3.60GHz
BenchmarkStringBase10BigInt/generic-8             154038              7792 ns/op            1024 B/op         32 allocs/op
BenchmarkStringBase10/generic-8                   587526              2013 ns/op               0 B/op          0 allocs/op
BenchmarkStringBase16/generic-8                 10830475               110.0 ns/op             0 B/op          0 allocs/op
```

```
name                          time/op
StringBase10BigInt/generic-8  7.77µs ± 0%
StringBase10/generic-8        2.01µs ± 0%
StringBase16/generic-8         110ns ± 0%
```


holiman:
Sorry, I missed that this PR existed. It is still marked wip, but I guess it's more or less done?

conversion.go
```diff
@@ -525,6 +528,29 @@ func (z *Int) Hex() string {
...
+		return dst.SetString(string(src), 0)
+	}
+
+	return fmt.Errorf("cannot scan %T", src)
+}
+
+// Value implements the database/sql/driver Valuer interface.
+// It encodes a string, because that is what postgres uses for its numeric type
+func (src Int) Value() (driver.Value, error) {
+	return string(src.ToBig().String()) + "e0", nil
```
holiman, line 551: Why `string(..)` on a `string` ? And why the concatenation of `e0`?

elee1766:
@holiman no problem, thanks for taking the time.

yeah, it's marked as WIP because I am not exactly sure if this is what everyone prefers (in terms of the sql stuff), and performance isn't super optimized.

if you're fine with that for now, I can mark as done. otherwise let me know if there is anything you would like me to change

conversion.go
```diff
@@ -525,6 +528,29 @@ func (z *Int) Hex() string {
...
+	case string:
+		return dst.SetString(src, 0)
+	case []byte:
+		return dst.SetString(string(src), 0)
+	}
+
+	return fmt.Errorf("cannot scan %T", src)
+}
+
+// Value implements the database/sql/driver Valuer interface.
```
holiman, line 548: We could add some compile time interface checks, like 
```
var (
	_ driver.Value   = (*Int)(nil)
	_ sql.Scanner   = (*Int)(nil)
	_ encoding.TextMarshaler = (*Int)(nil)
	_ encoding.TextUnmarshaler  = (*Int)(nil)
	_ json.Marshaler             = (*Int)(nil)
	_ json.Unmarshaler           = (*Int)(nil)
)
```

holiman (commented):
Looking ok. I think we should also add a fuzzer which compares the `SetString` in `Int` against `big.Int` `SetString`.

conversion.go
elee1766: no clue why i cast a string to string, will fix lol.

The e0 is to ensure that a large number is properly passed and understood as as "numeric/decimal". postgres wasn't happy when i fed it very large base 10 integer strings without the e0. postgres is able to downcast integers with exponential notation that are below size int64 to bigint transparently, so this felt like the most inclusive solution.

https://www.postgresql.org/docs/current/sql-syntax-lexical.html section 4.1.2.6. Numeric Constants


that said, I don't use any SQL-style DB's that aren't postgresql compatible. Not sure how MariaDB/Spanner will treat this

elee1766:
@holiman 

i've added a fuzz test for base10 using go's built in fuzzer. It runs when go test runs.

It checks that 
1. fail when SetString fails 
2. fail when number is negative 
3. fail when n umber is too large
4. succeeds otherwise
5. is equal in value to the big.Int when ToInt().Cmp is used
6. the base 10 strings 10 match
7. the valuer can Value() when the base10 string is valid
8. the Value() is equal to the base 10 string with an "e0" suffix

I've also changed SetString to closer match the big.Int interface, and changed the errors returned by FromBase10

conversion.go
```diff
@@ -98,6 +112,9 @@ func (z *Int) UnmarshalText(input []byte) error {
 // SetFromBig converts a big.Int to Int and sets the value to z.
 // TODO: Ensure we have sufficient testing, esp for negative bigints.
 func (z *Int) SetFromBig(b *big.Int) bool {
+	if b == nil {
+		return true
+	}
```
holiman, line 117: Seems weird to `no-op` here. IMO we should either crash or _at least_ make sure to clear `z`. 

But I don't know, seems like the caller should do the `nil`-check.

elee1766: hm, i agree the caller should do the nil check.

regardless i think i will leave it in for convenience, and just make sure to clear z

conversion.go
```diff
@@ -99,6 +113,9 @@ func (z *Int) UnmarshalText(input []byte) error {
 // TODO: Ensure we have sufficient testing, esp for negative bigints.
 func (z *Int) SetFromBig(b *big.Int) bool {
 	z.Clear()
+	if b == nil {
+		return true
+	}
```
holiman, line 118: In order to get a verdict, I checked what `big` does: 
```
func TestFoobar(t *testing.T) {
	new(big.Int).Set(nil)
}
```
==>
```
--- FAIL: TestFoobar (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
```

So, this is just defensive programming, let's remove it.

elee1766: cool. just removed.

holiman:
Sorry for being so slow on this, and thanks for keeping it up! 

Regarding the API, with this PR, it becomes

- Public "convenience" constructors
  - `FromBig(*big.Int)`
  - `FromHex(string)`
- Instance setters: 
  - `(z *Int) FromBase10(string)`
  - `(z *Int) SetFromBig(*big.Int)` 
  - `(z *Int) SetBytes([]byte)` 
  - `(z *Int) SetString(string, int)` 

And  a few more. The odd one out is `FromBase10`, being an instance-method. I think it would be more consistent to have 


- Public "convenience" constructors
  - `FromBig(*big.Int)`
  - `FromHex(string)`
  - `FromBase10(string)`
- Instance setters: 
  - `(z *Int) SetFromBase10(string)` (possibly)
  - `(z *Int) SetFromBig(*big.Int)` 
  - `(z *Int) SetBytes([]byte)` 
  - `(z *Int) SetString(string, int)`

holiman:
I pushed a minor change to make the API more coherent

base10.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (z *Int) fromBase10Long(bs string) error {
+	z.Clear()
+	if bs == "" {
+		return nil
+	}
+	iv := 19
+	c := 0
+	if len(bs) >= (iv * 4) {
```
holiman, line 88: Could you add a bit of description what happens inside this method -- what algorithm you're following? 
Doesn't have to be a lot, but a brief description. 
Also, what does `iv` mean? I associate the term with cryptographic  input vector, but this is probably something else? 
Also, is the unrolled loop really worth it? If there's no tangible gain, I'd prefer to have it in a shorter form

base10_test.go
```diff
@@ -0,0 +1,266 @@
...
+		// if its negative number, we should err
+		if len(orig) > 0 && (orig[0] == '-') {
+			if !errors.Is(err, ErrSyntaxBase10) {
+				t.Errorf("should have errored at negative number: %s", orig)
+			}
+			return
+		}
+		// if its too large, ignore it also
+		if val.Cmp(max256.ToBig()) > 0 {
+			return
```
holiman, line 111: If >256, we should have a non-nil `err`. Should check that

base10.go
gfxlabs: lol. i think iv originally meant "index value" or something like that.

I've cleaned up the function and added some comments.  - let me know if it's still unclear

base10_test.go
gfxlabs: added this

base10.go
```diff
@@ -0,0 +1,132 @@
...
+	// one less means that a string of 19 9's is always within the uint64 limit
+	cutLength := 19
+	// cutStart tracks the current position of the string that we are in
+	cutStart := 0
+	// start iterating from 4 to 1. This is because the maximum value of uint256 is 78 characters,
+	// which can be divided into 5 integers of up to 19 characters.
+	// however, the last number will always be below 19 characters, so i=0 is dealt with as special case
+	for i := 4; i >= 1; i-- {
+		// check if the length of the string is larger than cutLength * i
+		if len(bs) >= (cutLength * i) {
```
holiman, line 101: Reverse the check, so you can unindent the clause
```
if len(bs) < (cutLength * i){
    continue
}
```
Although, I wonder if it wouldn't make sense to start a better index. Something like this (might be off-by-one):
```
for i := len(bs) / 19; ...
```
Also, cutLength is a constant. It's ok to have it as a `var`, but I don't know, we could just also inline `19`. I don't know which I prefer.

elee1766: good catch on the loop, my bad! switched it to use continue and start at len(bs)/cutLength

I don't think it's off by one since if it's below 19, we would want it to be 0, and ignore the loop altogether, while otherwise, we are looking for the floor. if it's 19, it will perfectly match, and the stragglers case will get short circuited

cutLength is only used in this function - to me it makes more sense to leave it inside the function, unless other things start reusing that number 19.

base10.go
```diff
@@ -0,0 +1,132 @@
...
+
+var scaleTable10 [78]Int
+
+func init() {
+	for k := range scaleTable10 {
+		scaleTable10[k].Exp(NewInt(10), NewInt(uint64(k)))
+	}
+}
+
+// helper function to only ever be called via SetFromBase10
```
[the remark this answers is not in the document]
elee1766: > All in all, it's a pretty neat algo. I haven't ever dived into this problem before, it's interesting to see a clever algo for what actually is a trickier problem than one might first assume

admittedly, it was the only solution I could think of that wouldn't require writing my own esoteric lookup tables. (aka im lazy and am using the ones provided by the well optimized SetUint function)

luckily, we are only using this function when scanning from database, or parsing user input. If i had more performance sensitive use cases, I would maybe spend the time to sit down and make it better, for now, i think having no alloc is more than enough.

holiman:
Committed some failing tests now. Seems like this is a pretty gnarly thing to try to get equality with `big.Int` about, esp re base `0`. 

```
// SetString sets z to the value of s, interpreted in the given base,
// and returns z and a boolean indicating success. The entire string
// (not just a prefix) must be valid for success. If SetString fails,
// the value of z is undefined but the returned value is nil.
//
// The base argument must be 0 or a value between 2 and MaxBase.
// For base 0, the number prefix determines the actual base: A prefix of
// “0b” or “0B” selects base 2, “0”, “0o” or “0O” selects base 8,
// and “0x” or “0X” selects base 16. Otherwise, the selected base is 10
// and no prefix is accepted.
//
// For bases <= 36, lower and upper case letters are considered the same:
// The letters 'a' to 'z' and 'A' to 'Z' represent digit values 10 to 35.
// For bases > 36, the upper case letters 'A' to 'Z' represent the digit
// values 36 to 61.
//
// For base 0, an underscore character “_” may appear between a base
// prefix and an adjacent digit, and between successive digits; such
// underscores do not change the value of the number.
// Incorrect placement of underscores is reported as an error if there
// are no other errors. If base != 0, underscores are not recognized
// and act like any other character that is not a valid digit.

```
So we _should_ also support underscores, like `100_000_000`. And the docs doesn't even mention sign-characters that needs to be lopped off _before_ considering the prefix 
```
func TestFoo(t *testing.T) {
	s := "+0b00000000000000000000000000000000000000000000000000000000000000010"
	b, ok := new(big.Int).SetString(s, 0)
	fmt.Printf("b: %v, ok : %v\n", b, ok)
	z, ok := new(Int).SetString(s, 0)
	fmt.Printf("z: %v, ok : %v\n", z, ok)
}
```
^ The big.Int SetString accepts this, but we currently do not. 

Starting to wonder if maybe we should scale down, somehow, and implement a smaller subset that can be used for certain convenience situations, but do not map 1:1 to big.Int methods.

holiman:
> Starting to wonder if maybe we should scale down, somehow, and implement a smaller subset that can be used for certain convenience situations, but do not map 1:1 to big.Int methods.

Maybe not do `SetString` at all (since it by implication should be identical to `big.Int` setstring), but keep `..FromDecimal`

holiman:
I removed the SetString now. WDYT?

elee1766:
> I removed the SetString now. WDYT?

yeah the SetString api is a nightmare lmao. 

keeping FromDecimal and gutting SetString until a 1-1 match can be created seems like the right move.

 I am currently using SetString in one of my applications - but i can easily re implement it as SetString(uint256.Int, string) on the application side. i think you're 100% right that its implied that the implementation should match

conversion.go
```diff
@@ -51,12 +65,24 @@ func FromBig(b *big.Int) (*Int, bool) {
 	return z, overflow
 }
 
+// SetFromDecimal sets z from the given string, interpreted as a decimal number.
```
holiman, line 68:
```suggestion
// SetFromHex sets z from the given string, interpreted as a hexadecimal number.
```

decimal.go
```diff
@@ -0,0 +1,117 @@
...
+		if err != nil {
+			return err
+		}
+		// add that number to our running total
+		if i != 0 {
+			base := NewInt(num)
+			z.Add(z, base.Mul(base, mult))
+		} else {
+			z.SetUint64(num)
+		}
```
holiman, line 109:
```suggestion
		if i == 0 {
			z.SetUint64(num)
		} else {
			base := NewInt(num)
			z.Add(z, base.Mul(base, mult))
		}
```

conversion.go
```diff
@@ -24,6 +28,16 @@ const (
 	_ uint = -(maxWords & ^(4 | 8))       // maxWords is 4 or 8.
 )
 
+// Compile time interface checks
+var (
+	_ driver.Value             = (*Int)(nil)
```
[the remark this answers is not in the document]
holiman: This comment is old, I thought I had added it earlier. The `Value`/`Valuer` I have fixed already, but the remaining is still relevant. 

TL;DR: I think the `Value` and `Scan` are a bit underspecified. Hard to say if they are "correct" when there's no spec to match them against.

elee1766:
> This comment is old, I thought I had added it earlier. The `Value`/`Valuer` I have fixed already, but the remaining is still relevant.
> 
> TL;DR: I think the `Value` and `Scan` are a bit underspecified. Hard to say if they are "correct" when there's no spec to match them against.

moving out to main thread because i think its important, and maybe if anyone else has an opinion, easier for them to see and chime in.

so taking a step back, "Value" and "Scan" are for reading out of SQL databases. I added them since we have a need to store uint256 values in a postgres database for our uniswap foundation project.  

Value should return a valid on wire format for writing to an sql database, while Scan should parse a valid on wire format when reading from sql database 

so imo, there are three major dialects to consider, that is 

mysql  (mysql, mariadb, spanner)
psql  (postgres, cockroachdb, alloy)
sqlite  (sqlite, dq/rqlite)

there's also mssql, but fuck microsoft. 

sqlite has no 'numeric' datatype
psql has a numeric datatype: https://www.postgresql.org/docs/9.5/datatype-numeric.html 
mysql has a numeric datatype: https://dev.mysql.com/doc/refman/8.0/en/fixed-point-types.html

so for Value()

both [PSQL](https://www.postgresql.org/docs/9.3/datatype-numeric.html) and [MYSQL](https://dev.mysql.com/doc/refman/5.6/en/number-literals.html) support the scientific notation numeric literal (ala "XeY"). I actually can't find anywhere in the documentation postgresql talks about it, I only know since its in the lexer postgres and cockroach both use. We use postgres and have been writing with the e0. For some reason i remember having an issue when not writing the e0, but that might have something to do with our database. Let me try to figure out why... 

for Scan()

PSQL will just send the base 10 ascii bytes of the number, i assume mysql does the same(i dont know), but that's why it's written the way it is.


I think ideally, Value() rendering the ascii base 10 string without the e0 is best, since it would work in sqlite (store as varchar).

that said, i think postgresql compatibility is the most important of the three.  thegraph & cowswap off the top of my head both use postgres. 


I'll have some time later this weekend - i'll put together a quick SQL test suite with some different implementations and column types and see if we can switch back to ascii base 10

elee1766:
ok, so i tested stuff out. 

in postgresql, it works without the e0, I don't know why it wasn't before, i suspect it was on my side. 
In mariadb/mysql, it works fine as well, but decimal only supports up to 65 digits (not enough for uint256). I guess people should use CHAR/VARCHAR. technically there's Bit[] but i don't really know how it works
in sqlite, can just use the base 10 string with TEXT

the hex encoding won't be able to be parsed by postgres or mysql into the numeric/decimal type, so i think that base10 string is best. i updated the pr to remove the e0

elee1766:
i was thinking about this a little bit.

@holiman what do you feel about parsing strings with "-" prefix as the two's complement? (so remove the -, parse the number, then Neg it)

it would be convenient, and i don't feel it is out of scope for the package. (it already supports sdiv, smod, neg, abs, etc)

of course, there is no way to tell if a string should be marshaled into negative or not, so i think it should always marshal to string as positive, but being able to parse negative string input doesn't seem like a super awful idea.

holiman:
> @holiman what do you feel about parsing strings with "-" prefix as the two's complement? (so remove the -, parse the number, then Neg it)

I'm not keen on it. `uint256` is "natively" unsigned. Yes, there are some ways to work with "negative" numbers, but those are clearly "special cases". I don't think we should parse negative numbers, it just adds complexity. 

Also, I think encoders/decoders should come in pairs, and as you say, "there is no way to tell if a string should be marshaled into negative or not, so i think it should always marshal to string as positive" -- I take that as a sign that it's not a good fit. 


Also, I saw that `big.Int` also implements a `Scanner`, namely the `fmt.Scanner` interface: 
```
// Scanner is implemented by any value that has a Scan method, which scans
// the input for the representation of a value and stores the result in the
// receiver, which must be a pointer to be useful. The Scan method is called
// for any argument to Scan, Scanf, or Scanln that implements it.
type Scanner interface {
	Scan(state ScanState, verb rune) error
}
```
So if we merge this PR, then we can never implement that interface, because they are mutually exclusive.... :/ But I think that's ok, just wanted to highlight it

holiman:
The coverage is now back at 100%, and I changed the circle-ci fuzzer to use golang native fuzzing. TODO there is to make our old fuzzer (fuzzTernaryData etc) compatible with the native format, so it's run on circle. 

And then to make them all work with oss-fuzz, and we can scrap the dvyakov fuzzers. But that's not for this PR. 


@chfast want to take a look before we merge this?

holiman:
@chfast also, I've done something wrong re the fuzz corpus cache, https://app.circleci.com/pipelines/github/holiman/uint256/399/workflows/050524d7-5ac0-4342-8764-e88ef5e73dce/jobs/1196?invite=true#step-109-1

elee1766:
> > @holiman what do you feel about parsing strings with "-" prefix as the two's complement? (so remove the -, parse the number, then Neg it)
> 
> I'm not keen on it. `uint256` is "natively" unsigned. Yes, there are some ways to work with "negative" numbers, but those are clearly "special cases". I don't think we should parse negative numbers, it just adds complexity.
> 
> Also, I think encoders/decoders should come in pairs, and as you say, "there is no way to tell if a string should be marshaled into negative or not, so i think it should always marshal to string as positive" -- I take that as a sign that it's not a good fit.

good points - totally agree. I think where negatives are needed - application level can deal with it. (i went through our codebase, it was only 3 places of, so not a big deal)

conversion.go
```diff
@@ -565,7 +566,25 @@ func (dst *Int) Scan(src interface{}) error {
...
+		if len(splt) < 2 {
+			return dst.SetFromDecimal(src)
+		}
+		err := dst.SetFromDecimal(splt[0])
+		if err != nil {
+			return err
+		}
+		if splt[1] == "0" {
+			return nil
+		}
```
holiman, line 579: `1e0` should be `1`, not `0` ... ? 
Not sure what `0e0` is defined as, need to check that.

elee1766: isn't that what this does?

In the previous lines, i do dst.SetFromDecimal(splt[0])

then if `splt[1] == 0`, then i just return, since the value of the string is just the part in front of the "e"

holiman: My bad, yes you're right

conversion.go
```diff
@@ -565,7 +566,25 @@ func (dst *Int) Scan(src interface{}) error {
...
+		exp := new(Int)
+		err = exp.SetFromDecimal(splt[1])
+		if err != nil {
+			return err
+		}
+		exp.Exp(NewInt(10), exp)
+		dst.Mul(dst, exp)
```
holiman, line 586: For `Exp` we have no overflow checks, 
```
// Exp sets z = base**exponent mod 2**256, and returns z.
```

For `Mul`, we do have `MulOverflow`, so we could check that and return error if it does not fit. But it feels a bit wonky to just silently fail the conversion (because of the `Exp`) without signalling an error. 

IMO, it would have been ok if the failure-mode had been: 
- The result is `X mod 2^256`, where `X` is the correct result. 

But I think the possible failures are different .. (?)

elee1766: i think you are right here, they will be different.

i think the popular postgresql scanner will error anyways if the value is too large to be scanned in, so I think that it is fine to do the same.

holiman:
I am not really fond of the whole scientific notation-parsing in `Scan`, but OTOH it's isolated into a new method, and most callers won't use it (unless they strictly need it for database reasons). So I can live with that there.

PR #126, opened 2023-02-24
uint256: implement fastssz marshaling interfaces

karalabe:
These methods should enable `uint256.Int` to be used in SSZ encoders (the of `fastssz` flavor at least).


conversion.go
```diff
@@ -465,6 +465,46 @@ func bigEndianUint56(b []byte) uint64 {
...
+// SizeSSZ implements the fastssz.Marshaler interface and returns the byte size
+// of the 256 bit int.
+func (z *Int) SizeSSZ() int {
+	return 32
+}
+
+// UnmarshalSSZ implements the fastssz.Unmarshaler interface and parses an encoded
+// integer into the local struct.
+func (z *Int) UnmarshalSSZ(buf []byte) error {
+	if len(buf) != 32 {
```
holiman, line 497: Any particular reason to not do like this? 
```suggestion
	if len(buf) < 32 {
```

conversion.go
```diff
@@ -465,6 +465,46 @@ func bigEndianUint56(b []byte) uint64 {
 		uint64(b[2])<<32 | uint64(b[1])<<40 | uint64(b[0])<<48
 }
 
+// MarshalSSZTo implements the fastssz.Marshaler interface and serializes the
+// integer into an already pre-allocated buffer.
+func (z *Int) MarshalSSZTo(dst []byte) ([]byte, error) {
+	if len(dst) != 32 {
+		return nil, fmt.Errorf("%w: have %d, want %d bytes", ErrBadBufferLength, len(dst), 32)
+	}
```
holiman, line 473:
```suggestion
	if len(dst) < 32 {
		return nil, fmt.Errorf("%w: have %d, want >= %d bytes", ErrBadBufferLength, len(dst), 32)
	}
```

conversion.go
```diff
@@ -465,6 +465,46 @@ func bigEndianUint56(b []byte) uint64 {
...
+
+// MarshalSSZ implements the fastssz.Marshaler interface and returns the integer
+// marshalled into a newly allocated byte slice.
+func (z *Int) MarshalSSZ() ([]byte, error) {
+	return z.MarshalSSZTo(make([]byte, 32))
+}
+
+// SizeSSZ implements the fastssz.Marshaler interface and returns the byte size
+// of the 256 bit int.
+func (z *Int) SizeSSZ() int {
```
holiman, line 490:
```suggestion
func (*Int) SizeSSZ() int {
```

conversion.go
karalabe: That would accept and ignore junk at the end. I'd guess we want to enforce struct parsing.

conversion.go
karalabe: That would accept and ignore junk at the end. I'd guess we want to enforce struct parsing.

conversion.go
```diff
@@ -465,6 +465,56 @@ func bigEndianUint56(b []byte) uint64 {
...
+	binary.LittleEndian.PutUint64(dst[24:32], z[3])
+
+	return dst[32:], nil
+}
+
+// MarshalSSZ implements the fastssz.Marshaler interface and returns the integer
+// marshalled into a newly allocated byte slice.
+func (z *Int) MarshalSSZ() ([]byte, error) {
+	blob := make([]byte, 32)
+	z.MarshalSSZTo(blob) // ignore error, cannot fail, surely have 32 byte space in blob
```
holiman, line 486:
```suggestion
	_, _ = z.MarshalSSZTo(blob) // ignore error, cannot fail, surely have 32 byte space in blob
```

holiman (approved):
LGTM

PR #128, opened 2023-03-20
add `MustFromBig` for nicer struct initialization

karalabe:
This PR allows the second form (last two lines) along side the current form (first two lines) of initializing new ints from bigs.

```go
var (
	costCap, _   = uint256.FromBig(item.Tx.Cost())
	gasTipCap, _ = uint256.FromBig(item.Tx.GasTipCap())
)
meta := &blobTxMeta{
	id:            id,
	size:          size,
	nonce:         item.Tx.Nonce(),
	costCap:       costCap,
	gasTipCap:     gasTipCap,
	gasFeeCap:     uint256.MustFromBig(item.Tx.GasFeeCap()),
	blobGasFeeCap: uint256.MustFromBig(item.Tx.BlobGasFeeCap()),
}
```

Panics can only be captured when a goroutine is unwinding, hence why the messy construct in the tests. Also, it would have been a tad cleaner to wait on the panic with `sync.WaitGroup` vs a channel, but the channel approach introducing a dependency on `sync`.

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -66,6 +66,17 @@ func FromBig(b *big.Int) (*Int, bool) {
 	return z, overflow
 }
 
+// MustFromBig is a convenience-constructor from big.Int.
+// Returns a new Int and panics if overflow occurred.
+func MustFromBig(b *big.Int) *Int {
+	z := &Int{}
+	overflow := z.SetFromBig(b)
+	if overflow {
+		panic("overflow")
+	}
+	return z
+}
+
 // SetFromHex sets z from the given string, interpreted as a hexadecimal number.
 // OBS! This method is _not_ strictly identical to the (*big.Int).SetString(..., 16) method.
 // Notable differences:
--- a/conversion.go
+++ b/conversion.go
@@ -70,7 +70,7 @@ func FromBig(b *big.Int) (*Int, bool) {
 // Returns a new Int and panics if overflow occurred.
 func MustFromBig(b *big.Int) *Int {
 	z := &Int{}
-	if overflow := z.SetFromBig(b); overflow {
+	if z.SetFromBig(b) {
 		panic("overflow")
 	}
 	return z
--- a/conversion.go
+++ b/conversion.go
@@ -70,7 +70,7 @@ func FromBig(b *big.Int) (*Int, bool) {
 // Returns a new Int and panics if overflow occurred.
 func MustFromBig(b *big.Int) *Int {
 	z := &Int{}
-	if overflow := z.SetFromBig(b) {
+	if overflow := z.SetFromBig(b); overflow {
 		panic("overflow")
 	}
 	return z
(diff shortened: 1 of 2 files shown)


conversion.go
```diff
@@ -66,6 +66,17 @@ func FromBig(b *big.Int) (*Int, bool) {
...
 }
 
+// MustFromBig is a convenience-constructor from big.Int.
+// Returns a new Int and panics if overflow occurred.
+func MustFromBig(b *big.Int) *Int {
+	z := &Int{}
+	overflow := z.SetFromBig(b)
+	if overflow {
+		panic("overflow")
+	}
```
holiman, line 76:
```suggestion
	if overflow := z.SetFromBig(b) {
		panic("overflow")
	}
```

conversion.go
```diff
@@ -66,6 +66,16 @@ func FromBig(b *big.Int) (*Int, bool) {
 	return z, overflow
 }
 
+// MustFromBig is a convenience-constructor from big.Int.
+// Returns a new Int and panics if overflow occurred.
+func MustFromBig(b *big.Int) *Int {
+	z := &Int{}
+	if overflow := z.SetFromBig(b) {
```
holiman, line 73:
```suggestion
	if overflow := z.SetFromBig(b); overflow {
```

karalabe:
Actually, might have been nicer to just call FromBig and handle the return, more future proof.

holiman:
> Actually, might have been nicer to just call FromBig and handle the return, more future proof.

You mean in this PR or from your side?

holiman:
Actually, I'm thinking that if you're in this case: 
```
gasFeeCap:     uint256.MustFromBig(item.Tx.GasFeeCap()),
```
And let's say we know nothing about `item.Tx` at this point. Then it would be better to use a method which silently just overflows, not overflows+panic. Because if this happens early in the stack before the tx is properly validated, it is better to obtain a "wrong sender" than crashing the app. 

So wouldn't you rather have a `uint256.TryFromBig` or something like that?

conversion.go
```diff
@@ -66,6 +66,16 @@ func FromBig(b *big.Int) (*Int, bool) {
 	return z, overflow
 }
 
+// MustFromBig is a convenience-constructor from big.Int.
+// Returns a new Int and panics if overflow occurred.
+func MustFromBig(b *big.Int) *Int {
+	z := &Int{}
+	if overflow := z.SetFromBig(b); overflow {
+		panic("overflow")
+	}
```
holiman, line 75:
```suggestion
	if z.SetFromBig(b) {
		panic("overflow")
	}
```

karalabe:
>So wouldn't you rather have a uint256.TryFromBig or something like that?

Not really, all these data are already parsed from disk or network into uint256, only the tx interfaces are still using big.Int and I don't want to overhaul all the interfaces for the blob pool.

holiman (approved):
LGTM

Issue #123, opened 2022-09-28
implement `log2`, `log10`, `log256` (and possibly `ln`)


diwu1989:
Can there be a natural log or base 2 log implemented?
Thanks

A-ndy-git:
+1. Also been looking for this functionality for some time.

I'm unsure of the best method to implement this?

elee1766:
there are a few different solidity libraries that implement logarithm. off the top of my head, the big name i know is oz, they have log2,log10,log256 in their math libraries: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/Math.sol 
log10 i think is used in their word -> ascii string function.

it may be desirable to replicate the results of one such library (not neccesarily oz's), since I think mirroring or performing on-chain operations is a common use case for this package.

diwu1989:
+1 yes, that was my reasoning for requesting this, because I wanted to match some of the LOG math used in smart contracts

elee1766:
> +1 yes, that was my reasoning for requesting this, because I wanted to match some of the LOG math used in smart contracts

happy to do some work on this. personally we use log2

which LOG math functions do you have immediate needs for? i can start with those and add more down the line

log2? log10? log256?

diwu1989:
Log2 is enough for UniswapV3 tickmath:
https://github.com/Uniswap/v3-core/blob/main/contracts/libraries/TickMath.sol

holiman:
Is there any difference between `Log2()` and `BitLen()`? Because the latter exists already...

karalabe:
> Is there any difference between `Log2()` and `BitLen()`? Because the latter exists already...

Yes, `BitLen` is `int(log2)`. But if you want to calculate `logB(N)`, you need `log2(N)/log2(B)` with full float representation, not truncated to int.

karalabe:
FWIW, here's a set of log implementation for big.Int https://github.com/robpike/ivy/blob/master/value/log.go

holiman:
Possible api-methods could be 
```
(z *Int) IntLog2( ... ) *Int
(z *Int) IntLog10( ... ) *Int

(z *Int) Log2( ... ) float64
(z *Int) Log10( ... ) float64
```
However, it would be simpler to just do 
```
(z *Int) IntLog2( ... ) *Int
(z *Int) IntLog10( ... ) *Int

(z *Int) Float64() float64

```

[...]

PR #132, opened 2023-03-23
implement `Float64() float64`

holiman:
Ref: https://github.com/holiman/uint256/issues/123 

This PR implements 
```
// Float64 returns the float64 value nearest to x.
//
// Note: The `big.Float` version of `Float64` also returns an 'Accuracy', indicating
// whether the value was too small or too large to be represented by a
// `float64`. However, the `uint256` type is unable to represent values
// out of scope (|x| < math.SmallestNonzeroFloat64 or |x| > math.MaxFloat64),
// therefore this method does not return any accuracy.
//
// OBS! Currently, this method is not optimized: but uses big.Int and big.Float.
func (z *Int) Float64() float64
```
Implementing `Float64` makes it possible to use e.g. the logarithm functions in the `math` package. 

Todo: 

- [x] Implement native to-float support without using package `big`

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -92,17 +92,16 @@ func (z *Int) Float64() float64 {
 	//
 	// 1: https://www.wikihow.com/Convert-a-Number-from-Decimal-to-IEEE-754-Floating-Point-Representation
 
-	const bias = uint64(1023) // double-precision uses 1023 as bias
+	bitlen := uint64(z.BitLen())
 
 	// Normalize the number, by shifting it so that the MSB is shifted out.
-	shifts := uint(1 + 256 - z.BitLen())
-
+	y := new(Int).Lsh(z, uint(1+256-bitlen))
 	// The number with the leading 1 shifted out is the fraction.
-	y := new(Int).Lsh(z, shifts)
 	fraction := y[3]
 
 	// The exp is calculated from the number of shifts, adjusted with the bias.
-	biased_exp := bias + uint64(256-shifts)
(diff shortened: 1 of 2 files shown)


holiman:
```
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i7-1270P
BenchmarkFloat64/Float64/uint256-8                246834              4681 ns/op               0 B/op          0 allocs/op
BenchmarkFloat64/Float64/uint256-8                564375              4651 ns/op               0 B/op          0 allocs/op
BenchmarkFloat64/Float64/uint256-8                256735              4545 ns/op               0 B/op          0 allocs/op
BenchmarkFloat64/Float64/uint256-8                567476              2117 ns/op               0 B/op          0 allocs/op
BenchmarkFloat64/Float64/uint256-8                562767              2114 ns/op               0 B/op          0 allocs/op
BenchmarkFloat64/Float64/big-8                     23724             55516 ns/op           23424 B/op        510 allocs/op
BenchmarkFloat64/Float64/big-8                     18283             64730 ns/op           23424 B/op        510 allocs/op
BenchmarkFloat64/Float64/big-8                     18610             60523 ns/op           23424 B/op        510 allocs/op
BenchmarkFloat64/Float64/big-8                     19528             62700 ns/op           23424 B/op        510 allocs/op
BenchmarkFloat64/Float64/big-8                     19828             56352 ns/op           23424 B/op        510 allocs/op

```

PR #134, opened 2023-03-28
implement `Log10` and `CmpUint64`

holiman:
Work in progress

Adds the following methods

```golang
// CmpUint64 compares z and x and returns:
//
//	-1 if z <  x
//	 0 if z == x
//	+1 if z >  x
func (z *Int) CmpUint64(n uint64) int 

// Log10 returns the log in base 10, floored to nearest integer.
// **OBS** This method returns '0' for '0', not `-Inf`.
func (z *Int) Log10() uint
```

### DIFF
--- a/uint256.go
+++ b/uint256.go
@@ -932,6 +932,21 @@ func (z *Int) Cmp(x *Int) (r int) {
 	return 1
 }
 
+// CmpUint64 compares z and x and returns:
+//
+//	-1 if z <  x
+//	 0 if z == x
+//	+1 if z >  x
+func (z *Int) CmpUint64(n uint64) int {
+	if z[0] > n || (z[1]|z[2]|z[3]) != 0 {
+		return 1
+	}
+	if z[0] == n {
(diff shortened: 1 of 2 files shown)


uint256.go
```diff
@@ -1302,3 +1317,95 @@ func (z *Int) Sqrt(x *Int) *Int {
...
+	// 10 ** 64
+	tenTo64 = &Int{0, 7942358959831785217, 16807427164405733357, 1593091}
+)
+
+// Log10 returns the log in base 10, floored to nearest integer.
+// **OBS** This method returns '0' for '0', not `-Inf`.
+func (z *Int) Log10() uint {
+	// For some bit-lengths, there's only one possible value. Example:
+	// three bits can only represent [100 ... 111], or [4 ... 7]
+	// Ergo, bitlen:4 -> log10 == 0
```
holiman, line 1335:
```suggestion
	// Ergo, bitlen:3 -> log10 == 0
```

PR #136, opened 2023-03-30
 implement `Log10` and `CmpUint64` (alternate version)

holiman:
Alternative to #134

This PR implements `Log10` in what is probably close to the optimal way. Both PRs handle the majority of cases by one lookup based on the bitlength. 

For the remaining cases, #134 does 0, or, 1, or 2 divisions, and then a few uint64-compares and then it's done. 

In this PR, for the remaining cases, we do an additional lookup, and then a comparison, and then we're done. 

The drawback of this PR is that it has more precalculated 'persistent memory' hog. 
```

pows = [60]Int         // 1920B 
pows64 = [20]uint64 // 20*8 = 160B
```
So around 2 K. Not much, but still


holiman:
From CI benchmark run

#134 : 
```
BenchmarkLog10/Log10/uint256-36                     	  168109	      6829 ns/op	       0 B/op	       0 allocs/op
BenchmarkLog10/Log10/big-36                         	   36321	     34610 ns/op	   23424 B/op	     510 allocs/op
```

This PR
```
BenchmarkLog10/Log10/uint256-36                     	  705776	      1597 ns/op	       0 B/op	       0 allocs/op
BenchmarkLog10/Log10/big-36                         	   27064	     44892 ns/op	   23424 B/op	     510 allocs/op
```
So this is on the order of 3-4x faster.

PR #137, opened 2023-03-31
Allow converting between nil big.Int and nil uint256.Int

karalabe:
This is perhaps a debatable change, but may be more logical than originally anticipated.

In some places, having a `nil` as a `big.Int` or `uint256.Int` might make sense (e.g. an optional and non-initialized field). In those cases, converting between the two across older/newer code might be necessary. In it's current incarnation, the uint256 library will crash, so all optional nils need to be special cased.

IMHO this should not be necessary. If a uint256 should not be nil, but a nil big int is passed, then I assume the code later will crash anyway on some uint256 accessor method whether we crash at the conversion or not, so it's not like we're silently swallowing up errors. On the other hand, allowing the nils to be converted back and forth reduces the error potential of forgetting to special case an optional nil somewhere and blowing up runtime.

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -42,6 +42,9 @@ var (
 
 // ToBig returns a big.Int version of z.
 func (z *Int) ToBig() *big.Int {
+	if z == nil {
+		return nil
+	}
 	b := new(big.Int)
 	switch maxWords { // Compile-time check.
 	case 4: // 64-bit architectures.
@@ -62,6 +65,9 @@ func (z *Int) ToBig() *big.Int {
 // FromBig is a convenience-constructor from big.Int.
 // Returns a new Int and whether overflow occurred.
 func FromBig(b *big.Int) (*Int, bool) {
+	if b == nil {
+		return nil, false
+	}
 	z := &Int{}
 	overflow := z.SetFromBig(b)
 	return z, overflow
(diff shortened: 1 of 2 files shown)


holiman (approved):
I agree

conversion_test.go
```diff
@@ -211,7 +223,10 @@ func TestFromBigOverflow(t *testing.T) {
 }
 
 func TestToBig(t *testing.T) {
-
+	var uint256Nil *Int
+	if bigNil := uint256Nil.ToBig(); bigNil != nil {
+		t.Errorf("expected big.Int <nil>, got %x", bigNil)
```
holiman, line 228:
```suggestion
		t.Errorf("want big.Int <nil>, have %x", bigNil)
```
nit just to trigger CI

conversion.go
```diff
@@ -42,6 +42,9 @@ var (
 
 // ToBig returns a big.Int version of z.
```
holiman, line 43:
```suggestion
// ToBig returns a big.Int version of z.
// Return `nil` if z is nil
```

conversion.go
```diff
@@ -62,6 +65,9 @@ func (z *Int) ToBig() *big.Int {
 // FromBig is a convenience-constructor from big.Int.
 // Returns a new Int and whether overflow occurred.
```
holiman, line 66:
```suggestion
// Returns a new Int and whether overflow occurred.
// OBS: If b is `nil`, this method returns `nil, false`
```

conversion.go
```diff
@@ -70,6 +76,9 @@ func FromBig(b *big.Int) (*Int, bool) {
 // MustFromBig is a convenience-constructor from big.Int.
 // Returns a new Int and panics if overflow occurred.
```
holiman, line 77:
```suggestion
// Returns a new Int and panics if overflow occurred.
// OBS: If b is `nil`, this method does _not_ panic, but 
// instead returns `nil`
```

holiman (changes requested):

Issue #135, opened 2023-03-29
Allow Cmp to big.Int without converting big.Int


wjmelements:
##### Feature
```
func (*uint256.Int) CmpBig(big.Int) int
```

##### Motivation
Generally uint256.Int is superior and should replace big.Int everywhere. But in forked or in sufficiently large codebases it may not be practical or necessary to replace all instances. At the boundaries, this function could circumvent an explicit conversion.

I didn't find this feature when I looked for it so please close this if it already exists.

PR #138, opened 2023-03-31
Implement `CmpBig`

holiman:
This PR implements
```
// CmpBig compares z and x and returns:
//
//	-1 if z <  x
//	 0 if z == x
//	+1 if z >  x
func (z *Int) CmpBig(x *big.Int) (r int) 
```
After some benchmarking, it turned out that converting from big to `Int` was not significantly slower than e.g pre-checking the `BitLength`, so I went with that in the end.  

Closes https://github.com/holiman/uint256/issues/135

### DIFF
--- a/benchmarks_test.go
+++ b/benchmarks_test.go
@@ -375,9 +375,32 @@ func benchmark_Cmp_Bit(bench *testing.B) {
 		f2.Cmp(f)
 	}
 }
+
 func BenchmarkCmp(bench *testing.B) {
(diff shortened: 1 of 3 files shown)


holiman:
From CI run
```
BenchmarkCmpBig/nonequal-36               	82800283	        16.19 ns/op	       0 B/op	       0 allocs/op
BenchmarkCmpBig/equal-36                  	126040651	        10.66 ns/op	       0 B/op	       0 allocs/op
```

PR #144, opened 2023-11-26
conversion: behavioural changes in `String`, `MarshalText` and `MarshalJSON`

holiman:
This PR changes the way marshalling and unmarshalling behaves, in order to maximize compatibilty with `big.Int`: as in, maximize the chance that a 'drop in' replacement of `big.Int` for `uint256.Int` will work seamlessly. 

- `String()` - before this change, `String()` would return the integer in hexadecimal format. This PR changes it to instead return the input in decimal format, like `big.Int` does. 
- `UnmarshalText` now accepts either hex, `0x234` or `234`. Previously it accepted _only_ hex.
- `UnmarshalJSON` now accepts either hex, `0x234` or `234`. Previously it accepted _only_ hex.

 JSON marshalling is, alas, not 100% compatible, see https://github.com/holiman/uint256/compare/marshalling_compat?expand=1#diff-4caa05d97eba477e577a7721860f68293d7371001a42747af73e647b581a5e9bR630


holiman:
cc @karalabe

conversion.go
```diff
@@ -613,26 +618,38 @@ func (z *Int) EncodeRLP(w io.Writer) error {
...
 
 // MarshalText implements encoding.TextMarshaler
+// MarshalText marshals using the decimal representation (compatible with big.Int)
 func (z *Int) MarshalText() ([]byte, error) {
-	return []byte(z.Hex()), nil
+	return []byte(z.Dec()), nil
 }
 
 // MarshalJSON implements json.Marshaler.
+// MarshalText marshals using the hexadecimal representation. This is _not_ compatible
```
namiloh, line 627:
```suggestion
// MarshalJSON marshals using the hexadecimal representation. This is _not_ compatible
```

conversion.go
```diff
@@ -613,26 +618,38 @@ func (z *Int) EncodeRLP(w io.Writer) error {
...
+// with big.Int: big.Int marshals into JSO 'native' numeric format.
+//
+// The JSON  native format is, on some platforms, (e.g. java), limited to 53-bit large
```
namiloh, line 630:
```suggestion
// The JSON  native format is, on some platforms, (e.g. javascript), limited to 53-bit large
```

conversion.go
```diff
@@ -147,6 +147,7 @@ func (z *Int) fromHex(hex string) error {
 	if len(hex) > 66 {
 		return ErrBig256Range
 	}
+	z.Clear()
```
karalabe, line 150: If you moved this Clean in, it becomes unnecessary on Line 138?

conversion.go
```diff
@@ -613,26 +618,38 @@ func (z *Int) EncodeRLP(w io.Writer) error {
...
 
 // MarshalJSON implements json.Marshaler.
+// MarshalJSON marshals using the hexadecimal representation. This is _not_ compatible
+// with big.Int: big.Int marshals into JSON 'native' numeric format.
+//
+// The JSON  native format is, on some platforms, (e.g. javascript), limited to 53-bit large
+// integer space. Thus, U256 uses string-format, which is not compatible with
+// big.int (big.Int refuses to unmarshal a string representation).
+// Since we cannot reach compatibility with big.Int _anyway_, we might aswell
+// choose to use hexadecimal representation here anyway.
```
karalabe, line 634: I am unsure about this logic.

I think "the standard" solution employed by most libraries hitting this issue is to return a decimal string. Hex seems very Ethereum specific and I feel this lib should not leak out those decisions at such a low level. Anything against the decimal formatting by default?

holiman: Changed to decimal

conversion_test.go
```diff
@@ -1216,38 +1218,68 @@ func TestEnDecode(t *testing.T) {
...
+			t.Fatalf("test %d Dec(), have %v, want %v", i, have, want)
 		}
 		{ // Json
 			jsonEncoded, err := json.Marshal(&jsonStruct{&intSample})
 			if err != nil {
-				t.Fatalf("test %d #6, err: %v", i, err)
+				t.Fatalf("test %d: json encoding err: %v", i, err)
 			}
+			jsonEncodedBig, _ := json.Marshal(&jsonBigStruct{&bigSample})
+			//if !bytes.Equal(jsonEncoded, jsonEncodedBig) {
```
karalabe, line 1261: Any specific reason for this big commented out chunk of code?

holiman: not really

PR #145, opened 2023-11-30
conversion: revert json-marshalling change in #144, marshal hex-format

holiman:
This undoes the change in json-marshalling, making it so that we (again) marshal `uint256.Int` in hex-format. This is more widely useful in go-ethereum. 

I originally intended to let it be this way, but I changed my mind due to this comment: https://github.com/holiman/uint256/pull/144#discussion_r1405828059. 

This would be an alternative to https://github.com/ethereum/go-ethereum/pull/28628

cc @fjl @karalabe

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -623,14 +623,14 @@ func (z *Int) MarshalText() ([]byte, error) {
 }
 
 // MarshalJSON implements json.Marshaler.
-// MarshalJSON marshals using the 'decimal string' representation. This is _not_ compatible
+// MarshalJSON marshals using 'hex' representation. This is _not_ compatible
 // with big.Int: big.Int marshals into JSON 'native' numeric format.
 //
 // The JSON  native format is, on some platforms, (e.g. javascript), limited to 53-bit large
 // integer space. Thus, U256 uses string-format, which is not compatible with
 // big.int (big.Int refuses to unmarshal a string representation).
 func (z *Int) MarshalJSON() ([]byte, error) {
-	return []byte(`"` + z.Dec() + `"`), nil
+	return []byte(`"` + z.Hex() + `"`), nil
 }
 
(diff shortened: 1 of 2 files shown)


karalabe:
But why?

holiman:
As @fjl noted: 

> However, let me just say I find the upstream change is not the best. We always want hex output in go-ethereum. So now we need to add a conversion every time just like for big.Int. I thought uint256 is purpose built for our use, why would we make it harder for ourselves?

lmittmann:
What is the status here? will this change from hex to decimal format be reverted or not? Having decimal marshaling is kind of annoying in the ethereum context and would require a proxy type as `hexutil.Big` everywhere.

holiman:
We discussed this, and agreed to not revert it. I implemented a hexutil.U256 type in go-ethereum, and closed this PR.

lmittmann:
Okay. Just out of curiosity, what is the reason for changing this in the first place?

To give some more context: My lib is effected by this breaking change (I presume it will be relatively easy to fix though). But still, this package has [hundreds of direct dependents](https://deps.dev/go/github.com%2Fholiman%2Fuint256) which code might break or act unexpected. This should be a v2 release IMO or kept as is.

holiman:
> Okay. Just out of curiosity, what is the reason for changing this in the first place?

To be more like `big.Int`. Or exactly which of the changes are you referring to? 


> To give some more context: My lib is effected by this breaking change (I presume it will be relatively easy to fix though). But still, this package has [hundreds of direct dependents](https://deps.dev/go/github.com%2Fholiman%2Fuint256) which code might break or act unexpected. This should be a v2 release IMO or kept as is.

Well, depends on how you see it. It doesn't break the programing API, but yes, it does break some places where marshalling is concerned. But many places where you would unmarshal a hex-string, would also accept an int-string. 

Anyway, go-ethereum is the main driver for this library.

PR #146, opened 2023-12-12
add global zero

own2pwn:


holiman:
Thanks but it's trivial for anyone to either make one on the spot (on the stack instead of on the heap), and also, a global "constant" which cannot be guaranteed to actually always _be_ constant is not awesome, IMO. I mean, any caller can mutate it to be non-zero.

PR #148, opened 2024-03-18
Chore: Reordered the `SetBytesXXX` function and fixed comments

notJoon:
## Description

I sorted the `SetBytesN` functions in ascending order of `N`, and corrected parts where the name in the comment was wrong. For example, the `SetBytes24` function had `SetBytes16` in its comment. Additionally, since only some functions had comments, I added comments to the rest based on those previous comments.

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -254,11 +254,129 @@ func (z *Int) Format(s fmt.State, ch rune) {
 	z.ToBig().Format(s, ch)
 }
 
+// SetBytes1 is identical to SetBytes(in[:1]), but panics is input is too short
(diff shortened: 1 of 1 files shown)


holiman (commented):
Thank you, but I prefer to keep it the way it is. Your way of ordering it makes sense from some kind of natural-order point of view. The reason it's ordered differently is for easier review. 

The SetBytes8,16,24 are all pretty similar. 
The Setbytes3,11,19 and 27 are also pretty similar. 

So, the way they are ordered makes it easier for anyone who wants to dive in and review. 

The docstring-additions were nice, but you also copy-pasted a typo `but panics is input is too short` so not quite good enough :)

PR #150, opened 2024-03-21
uint256: simplify code with `gofmt -s`

AaronChen0:
An array, slice, or map composite literal of the form:
[]T{T{}, T{}}
can be simplified to:
[]T{{}, {}}

This is one of the simplifications that "gofmt -s" applies.


holiman (approved):
LGTM, thanks

PR #154, opened 2024-03-31
uint256: optimize DivMod

AaronChen0:
Following Div and Mod, this commit adds shortcut trivial cases for DivMod.


holiman (approved):
LGTM, thanks!

PR #156, opened 2024-04-16
uint256: optimize leadingZeros and add BenchmarkBitLen

AaronChen0:
```
go test ./...
```
returns
```
ok  	github.com/holiman/uint256	0.977s
```

## Bechmark for `leadingZeros`
```
go test -run - -bench BenchmarkLead -benchmem
```
code:
```
func BenchmarkLeadingZeros(bench *testing.B) {
	// res is a variable to prevent the golang compiler from eliminating the benchmark testing entirely.
	var res int

	data := []Int{
		{0, 0, 0, 0xffffffffffffffff},
		{0, 0, 0xffffffffffffffff, 0},
		{0, 0xffffffffffffffff, 0, 0},
		{0xffffffffffffffff, 0, 0, 0},
	}

	benchmarkLeadingZeros := func(b *testing.B, num *Int) {
		var r int
		for i := 0; i < b.N; i++ {
			r = leadingZeros(num)
		}
		res = r
	}

	res += 1

	bench.Run("leadingZeros192/uint256", func(b *testing.B) { benchmarkLeadingZeros(b, &data[0])})
	bench.Run("leadingZeros128/uint256", func(b *testing.B) { benchmarkLeadingZeros(b, &data[1])})
	bench.Run("leadingZeros64/uint256", func(b *testing.B) { benchmarkLeadingZeros(b, &data[2])})
	bench.Run("leadingZeros0/uint256", func(b *testing.B) { benchmarkLeadingZeros(b, &data[3])})
}
```
old:
```
goos: linux
goarch: amd64
BenchmarkLeadingZeros/leadingZeros192/uint256-16         	238305988	         4.919 ns/op	       0 B/op	       0 allocs/op
BenchmarkLeadingZeros/leadingZeros128/uint256-16         	232537900	         4.886 ns/op	       0 B/op	       0 allocs/op
BenchmarkLeadingZeros/leadingZeros64/uint256-16          	244867848	         5.017 ns/op	       0 B/op	       0 allocs/op
BenchmarkLeadingZeros/leadingZeros0/uint256-16           	244553245	         5.197 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/holiman/uint256	6.845s
```
new:
```
goos: linux
goarch: amd64
BenchmarkLeadingZeros/leadingZeros192/uint256-16         	1000000000	         1.098 ns/op	       0 B/op	       0 allocs/op
BenchmarkLeadingZeros/leadingZeros128/uint256-16         	906194443	         1.316 ns/op	       0 B/op	       0 allocs/op
BenchmarkLeadingZeros/leadingZeros64/uint256-16          	779154236	         1.565 ns/op	       0 B/op	       0 allocs/op
BenchmarkLeadingZeros/leadingZeros0/uint256-16           	662557696	         1.786 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/holiman/uint256	5.301s
```

### DIFF
--- a/mod.go
+++ b/mod.go
@@ -10,13 +10,8 @@ import (
 
 // Some utility functions
 
-func leadingZeros(x *Int) (z int) {
-	var t int
-	z = bits.LeadingZeros64(x[3])
-	t = bits.LeadingZeros64(x[2]); if z ==  64 { z = t +  64 }
-	t = bits.LeadingZeros64(x[1]); if z == 128 { z = t + 128 }
-	t = bits.LeadingZeros64(x[0]); if z == 192 { z = t + 192 }
-	return z
+func leadingZeros(x *Int) int {
+	return 256 - x.BitLen()
 }
 
 // Reciprocal computes a 320-bit value representing 1/m
--- a/benchmarks_test.go
+++ b/benchmarks_test.go
@@ -944,3 +944,42 @@ func BenchmarkSet(bench *testing.B) {
 	bench.Run("single/uint256", benchmarkUint256)
 	bench.Run("single/big", benchmarkBig)
 }
+
+func BenchmarkBitLen(bench *testing.B) {
+	// res is a variable to prevent the golang compiler from eliminating the benchmark testing entirely.
+	var res int
+
+	data := []Int{
+		{0, 0, 0, 0xffffffffffffffff},
+		{0, 0, 0xffffffffffffffff, 0},
+		{0, 0xffffffffffffffff, 0, 0},
+		{0xffffffffffffffff, 0, 0, 0},
+	}
+
+	benchmarkBitLenUint256 := func(b *testing.B, num *Int) {
+		var r int
+		for i := 0; i < b.N; i++ {
+			r = num.BitLen()
+		}
+		res = r
+	}
+
+	benchmarkBitLenBig := func(b *testing.B, num *big.Int) {
+		var r int
+		for i := 0; i < b.N; i++ {
+			r = num.BitLen()
+		}
+		res = r
+	}
+
+	res += 1
+
+	bench.Run("BitLen192/uint256", func(b *testing.B) { benchmarkBitLenUint256(b, &data[0])})
+	bench.Run("BitLen128/uint256", func(b *testing.B) { benchmarkBitLenUint256(b, &data[1])})
+	bench.Run("BitLen64/uint256", func(b *testing.B) { benchmarkBitLenUint256(b, &data[2])})
+	bench.Run("BitLen0/uint256", func(b *testing.B) { benchmarkBitLenUint256(b, &data[3])})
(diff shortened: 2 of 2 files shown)


AaronChen0:
Hi. Can someone give some feedbacks on my recent PRs?

holiman:
Yep, I will look into it the coming week. Thanks for the PRs!

mod.go
```diff
@@ -10,13 +10,8 @@ import (
...
 // Some utility functions
 
-func leadingZeros(x *Int) (z int) {
-	var t int
-	z = bits.LeadingZeros64(x[3])
-	t = bits.LeadingZeros64(x[2]); if z ==  64 { z = t +  64 }
-	t = bits.LeadingZeros64(x[1]); if z == 128 { z = t + 128 }
-	t = bits.LeadingZeros64(x[0]); if z == 192 { z = t + 192 }
-	return z
+func leadingZeros(x *Int) int {
```
holiman, line 13: This method is not actually part of the 'functional' code, that is, code that is actually being used. It seems to be only used from `mod_test.go`, the method `TestLeadingZeroes`. 

So it's not worth optimizing, but since your implementation is a lot nicer, I think we should: 
 
- Not bother adding the new benchmark-tests (`BenchmarkBitLen`)
- Move `leadingZeroes` function into `mod_test.go`, to make it obvious that it's "supporting code" and not prod
- Make it use your new code

SGTY?

AaronChen0: I think we either remove `leadingZeroes` or make it public, since it's only used for `TestLeadingZeros`  in mod_test.go.
And BenchmarkBitLen is for benchmarking BitLen, not for leadingZeros.

AaronChen0: If we choose removing `leadingZeroes`, should I delete mod_test.go directly?

holiman: Oh right, yes, let's just nuke it

holiman: > And BenchmarkBitLen is for benchmarking BitLen, not for leadingZeros.

Ah, my bad. Then let's keep that, but I think we should perhaps, for each `N`, do a loop on all `0-256` bitsizes, instead of using just the four "magic" ones. Either that, or use random values.

AaronChen0: I think the 4 values can represent the 4 class of numbers well. And remember this is benchmarking, not testing.

holiman: Yes, but if we add a skewed benchmark, someone might submit a "faster" implementation which is faster in those special cases, but not in the general case. In general it's good if we make the benchmarks somewhat generic.

AaronChen0: Since we remove `leadingZeroes`, I am ok with not adding `BenchmarkBitLen`.

holiman: If we just use all bitsizes instead of only four (and clump it all into a single benchmark), that should be more or less a "perfect" sampling

PR #162, opened 2024-05-06
uint256: fix ExtendSign

AaronChen0:
Sorry. I miss a special case `z == byteNum`,  that is `z.ExtendSign(x,z)`.


holiman (approved):
LGTM, but would be nice to catch it in a testcase too

holiman:
> LGTM, but would be nice to catch it in a testcase too

No need to add it in this PR, I have a testing-PR coming which will add this

PR #167, opened 2024-05-14
uint256: minor improvement for Mul

AaronChen0:
Minor tweek of the code. Not sure where the boost comes from. But it does improve around 10% in my two x86-64 machines.

godbolt links:
old:  https://go.godbolt.org/z/z8vqhG9r7
new: https://go.godbolt.org/z/M17Ycfzxx
## Test
```
go test ./...
```
```
ok  	github.com/holiman/uint256	1.380s
```
## Benchmark
```
goos: linux
goarch: amd64
                      │     old     │                 new                 │
                      │   sec/op    │   sec/op     vs base                │
Mul/single/uint256-16   4.239n ± 2%   3.725n ± 1%  -12.14% (p=0.000 n=10)
Mul/single/big-16       37.05n ± 1%   36.96n ± 2%        ~ (p=0.403 n=10)
geomean                 12.53n        11.73n        -6.38%
```
```
goos: linux
goarch: amd64
cpu: Intel Core Processor (Broadwell, no TSX, IBRS)
                   │     old      │                new                 │
                   │    sec/op    │   sec/op     vs base               │
Mul/single/uint256   10.125n ± 4%   9.346n ± 2%  -7.69% (p=0.000 n=10)
Mul/single/big        79.24n ± 3%   79.11n ± 5%       ~ (p=0.853 n=10)
geomean               28.32n        27.19n       -4.00%
```

### DIFF
--- a/uint256.go
+++ b/uint256.go
@@ -367,25 +367,22 @@ func umul(x, y *Int, res *[8]uint64) {
 // Mul sets z to the product x*y
 func (z *Int) Mul(x, y *Int) *Int {
 	var (
-		carry                  uint64
-		res0, res1, res2, res3 uint64
+		carry0, carry1, carry2 uint64
+		res1, res2 uint64
+		x0, x1, x2, x3 = x[0], x[1], x[2], x[3]
+		y0, y1, y2, y3 = y[0], y[1], y[2], y[3]
 	)
 
-	carry, res0 = bits.Mul64(x[0], y[0])
-	carry, res1 = umulHop(carry, x[1], y[0])
-	carry, res2 = umulHop(carry, x[2], y[0])
-	res3 = x[3]*y[0] + carry
+	carry0, z[0] = bits.Mul64(x0, y0)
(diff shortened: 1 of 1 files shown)


holiman:
Interesting! 
```
name                  old time/op  new time/op  delta
Mul/single/uint256-8  15.1ns ± 0%  10.6ns ±35%  -29.68%  (p=0.000 n=8+10)
```
@chfast lgty?

chfast (approved):
LGTM as this only moves the values to different variables and changes the order of operations.

chfast:
> New: https://go.godbolt.org/z/5E8TahK9j Old: https://go.godbolt.org/z/nY31bsxdb
> 
> With Go 1.22.1, the new code has 7 less `MOVQ` instructions. Other instructions are the same amount. Possibly, this is the reason.

godbolt has also the "diff view": https://go.godbolt.org/z/ddcK6TzKK

It looks the new code spills less values from registers to stack. Good change.

Issue #170, opened 2024-05-20
MarshalSSZTo and other incompatibilities with fastssz


jshufro:
Fast ssz now expects `HashTreeRootWith(hh ssz.HashWalker) (err error)` in its `HashTreeRoot` interface. I believe the implementation is as simple as:
```golang
func (u *Uint256) HashTreeRootWith(hh ssz.HashWalker) (err error) {
        bytes := make([]byte, 32)
        bytes, err = u.MarshalSSZTo(bytes)
        if err != nil {
                return
        }

        hh.AppendBytes32(bytes)
        return
}
```
but I have been known to make mistakes.

Further, fastssz's `MarshalSSZTo(buf []byte) ([]byte, error)` interface function expects the implementation to append to the end of `buf`, not to overwrite it. Unfortunately, this library overwrites instead, and a wrapper is needed (one is needed anyway for sszgen to work):
```golang
type Uint256 struct {
        repr *uint256.Int `ssz:"-"`
}
[...]
func (u *Uint256) MarshalSSZTo(buf []byte) ([]byte, error) {
        bytes, err := u.repr.MarshalSSZ()
        if err != nil {
                return nil, err
        }
        return append(buf, bytes...), nil
}
```

holiman:
Thanks for the heads-up!

holiman:
> Fast ssz now expects `HashTreeRootWith(hh ssz.HashWalker) (err error)` in its `HashTreeRoot` interface. I believe the implementation is as simple as:

As for this, nothing we can do here. As the comment on `HashTreeRoot` says, https://github.com/holiman/uint256/blob/master/conversion.go#L585 , 
```
// HashTreeRoot implements the fastssz.HashRoot interface's non-dependent part.
```
That is, out of the methods in the `HashRoot` interface
```golang
type HashRoot interface {
	GetTree() (*Node, error)
	HashTreeRoot() ([32]byte, error)
	HashTreeRootWith(hh HashWalker) error
}
```
`HashTreeRot()` is the only one that can be implemented without introducing a dependency from `uint256` to `fastssz`.

jshufro:
You know, taking a step back, the whole situation is disappointing, as the interface _should_ expect a function that marshals ssz into a preallocated buffer. Maybe it does `make([]byte, 0, N)` to avoid reallocation, but that leaves you the unenviable implementation task of 1-copy serialization: first into a local buffer, then into the preallocated one.

[...]

PR #171, opened 2024-05-21
conversion: replace `MarshalSSZTo` with `MarshalSSZAppend` and `MarshalSSZInto`

holiman:
This fixes the append vs overwrite part of https://github.com/holiman/uint256/issues/170 .

However, this is technically a breaking change. On the other hand, the way it now works is incompatible with the fastssz encoding library. And since nobody has raised this earlier, I assume nobody has really used it in production.  

So, should be fine to merge, on that reasoning. 

(?)

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -540,25 +540,27 @@ func bigEndianUint56(b []byte) uint64 {
 		uint64(b[2])<<32 | uint64(b[1])<<40 | uint64(b[0])<<48
 }
 
-// MarshalSSZTo implements the fastssz.Marshaler interface and serializes the
-// integer into an already pre-allocated buffer.
+// MarshalSSZTo implements the fastssz.Marshaler interface, serializes and appends
+// the integer to the dst buffer, and returns the (potentially reallocated) buffer.
+//
+//	 https://github.com/ferranbt/fastssz/blob/main/interface.go#L4
+//		type Marshaler interface {
+//			MarshalSSZTo(dst []byte) ([]byte, error)
+//			MarshalSSZ() ([]byte, error)
+//			SizeSSZ() int
+//		}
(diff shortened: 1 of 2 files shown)


jshufro:
> (?)

I'd recommend going through the paces with deprecation instead, or at the very least keeping around a function that preserves the old behavior (under a new symbol).

holiman:
Deprecating `MarshalSSZTo` doesn't really work -- it's an interface method, so I think calls to to it are via not direct, and the deprecation won't be noticed by tooling.
 
Possibly, I could remove it, and add it back a few releases later. At least then the breakage would be obvious. And that's perhaps better than what this PR does, which breaks things in a way which will only show in runtime, not compile-time.

jshufro:
> calls to to it are via not direct

I can't imagine anyone is calling it any way except direct.

sszgen/fastssz do not really support external interfaces. The `-include` flag for sszgen expects a local package only. I had to do all this: https://github.com/jshufro/smartnode/blob/e55d271c981833425987fb84d20d03e6177a7276/shared/services/rewards/ssz_types/big/uint256.go

holiman:
```
type Uint256 struct {
	*big.Int
}
```
Hm, why can't you use `uint256.Int` directly, or, alternatively, alias it?
```golang
type Uint256 uint256.Int`
```

But I agree, perhaps best to just deprecate `MarshalSSZTo` in this PR, and create `MarshalSSZAppend` as a new method which does the append.

holiman:
> Possibly, I could remove it, and add it back a few releases later. At least then the breakage would be obvious.

I have now changed this PR to _remove_ `MarshalSSZTo`, to force users to choose either `MarshalSSZAppend` or `MarshalSSZInto`. The former is the future `MarshalSSZTo`, the latter is already deprecated. 

@jshufro SGTY?

PR #174, opened 2024-05-24
uint256: optimize Sqrt

AaronChen0:
1. Make `x.IsUint64()` a special case.
2. If `x.IsUint64()` is `false`, then 
```
z2.Div(x, z1)
```
 is equal to 
```
z2.Clear()
udivrem(z2[:], x[:], z1, nil)
```
3. `Lt` is faster than `Cmp`.
4. The first div can be replaced with a right shift. `Rsh` is 10x cheaper than `Div`.
## Benchmark
```
goos: linux
goarch: amd64
                       │     old     │                 new                 │
                       │   sec/op    │   sec/op     vs base                │
Sqrt/single/uint256-16   241.0n ± 1%   172.4n ± 1%  -28.46% (p=0.000 n=10)
Sqrt/single/big-16       994.1n ± 1%   988.9n ± 1%        ~ (p=0.089 n=10)
geomean                  489.5n        412.9n       -15.64%
```


uint256.go
```diff
@@ -1280,34 +1280,53 @@ func (z *Int) ExtendSign(x, byteNum *Int) *Int {
...
+			}
+			z1 = z2
+		}
 	}
-	var (
-		z1 = &Int{1, 0, 0, 0}
-		z2 = &Int{}
-	)
+
+	flag := false
```
holiman, line 1301: a flag called `flag` is not great for code clarity. Please rename to something expressing what it means, e.g. `first` or `notFirst`.

uint256.go
```diff
@@ -1280,34 +1280,53 @@ func (z *Int) ExtendSign(x, byteNum *Int) *Int {
...
-			z2[1] = (z2[1] >> 1) | b
-			z2[0] = (z2[0] >> 1) | a
+		// z2.Div(x, z1) -- x > MaxUint64, x > z1 > 0
+		if flag {
+			z2.Clear()
+			udivrem(z2[:], x[:], z1, nil)
+		} else {
+			// The first div is equal to a right shift
+			z2.Rsh(x, uint(x.BitLen()+1)/2)
+			flag = true
```
holiman, line 1315: I think you can move this outside the loop. So we get something like 
```golang
	// Start with value known to be too large and repeat "z = ⌊(z + ⌊x/z⌋)/2⌋" until it stops getting smaller.
	z1.Lsh(z1, uint(x.BitLen()+1)/2) // must be ≥ √x

	// We can do the first division outside the loop
	z2.Rsh(x, uint(x.BitLen()+1)/2) // The first div is equal to a right shift
	first := true
	for {
		// z2.Div(x, z1) -- x > MaxUint64, x > z1 > 0
		if !first { // first division was done outside the loop
			z2.Clear()
			udivrem(z2[:], x[:], z1, nil)
		}
		first = false
```

AaronChen0:
Follow your suggestion. Not sure why the DeepSource check timed out.

holiman:
We can simplify even more, by moving the division to the end of the loop, and lose the flag. 
For some reason, my linter wants to do further changes, but the CI-linter doesn't seem to complain. 
```diff
diff --git a/uint256.go b/uint256.go
index 493f44a..7b93ee9 100644
--- a/uint256.go
+++ b/uint256.go
@@ -369,9 +369,9 @@ func umul(x, y *Int, res *[8]uint64) {
 func (z *Int) Mul(x, y *Int) *Int {
 	var (
 		carry0, carry1, carry2 uint64
-		res1, res2 uint64
-		x0, x1, x2, x3 = x[0], x[1], x[2], x[3]
-		y0, y1, y2, y3 = y[0], y[1], y[2], y[3]
+		res1, res2             uint64
+		x0, x1, x2, x3         = x[0], x[1], x[2], x[3]
+		y0, y1, y2, y3         = y[0], y[1], y[2], y[3]
 	)
 
 	carry0, z[0] = bits.Mul64(x0, y0)
@@ -1283,14 +1283,14 @@ func (z *Int) Sqrt(x *Int) *Int {
 	if x.IsUint64() {
 		var (
 			x0 uint64 = x.Uint64()
-			z1 uint64 = 1 << ((bits.Len64(x0)+1)/2)
+			z1 uint64 = 1 << ((bits.Len64(x0) + 1) / 2)
 			z2 uint64
 		)
 		if x0 < 2 {
 			return z.SetUint64(x0)
 		}
 		for {
-			z2 = (z1 + x0 / z1) >> 1
+			z2 = (z1 + x0/z1) >> 1
 			if z2 >= z1 {
 				return z.SetUint64(z1)
 			}
@@ -1306,27 +1306,24 @@ func (z *Int) Sqrt(x *Int) *Int {
 
 	// We can do the first division outside the loop
 	z2.Rsh(x, uint(x.BitLen()+1)/2) // The first div is equal to a right shift
-	first := true
 
 	for {
-		// z2.Div(x, z1) -- x > MaxUint64, x > z1 > 0
-		if !first { // first division was done outside the loop
-			z2.Clear()
-			udivrem(z2[:], x[:], z1, nil)
-		}
-		first = false
 		z2.Add(z2, z1)
-		
+
 		// z2 = z2.Rsh(z2, 1) -- the code below does a 1-bit rsh faster
-		z2[0] = (z2[0] >> 1) | z2[1] << 63
-		z2[1] = (z2[1] >> 1) | z2[2] << 63
-		z2[2] = (z2[2] >> 1) | z2[3] << 63
+		z2[0] = (z2[0] >> 1) | z2[1]<<63
+		z2[1] = (z2[1] >> 1) | z2[2]<<63
+		z2[2] = (z2[2] >> 1) | z2[3]<<63
 		z2[3] >>= 1
 
 		if !z2.Lt(z1) {
 			return z.Set(z1)
 		}
 		z1.Set(z2)
+		// Next iteration of the loop
+		// z2.Div(x, z1) -- x > MaxUint64, x > z1 > 0
+		z2.Clear()
+		udivrem(z2[:], x[:], z1, nil)
 	}
 }
 
```

AaronChen0:
Good observation!

holiman:
```
goos: linux
goarch: amd64
cpu: 12th Gen Intel(R) Core(TM) i7-1270P
                      │ bench.sqrt.1 │             bench.sqrt.4             │
                      │    sec/op    │    sec/op     vs base                │
Sqrt/single/uint256-8   650.5n ± 42%   391.6n ± 11%  -39.80% (p=0.002 n=10)
```
:+1:

PR #179, opened 2024-07-25
uint256: fix an edge case in DivMod

### LINKED ISSUE #178: Small problem with DivMod
Hello everyone

I was kind of using DivMod, when I got an unexpected remainder 0 for the following
`z := uint256.MustFromDecimal("147120705406341664063220615392")`
`y := uint256.MustFromDecimal("1461501637330902918203679268441974103478767737394")`
`fmt.Println(z.DivMod(z, y, new(uint256.Int)))`

output: 0 0

Using new receiver returns the correct result.
`z = uint256.MustFromDecimal("147120705406341664063220615392")`
`y = uint256.MustFromDecimal("1461501637330902918203679268441974103478767737394")`
`fmt.Println(new(uint256.Int).DivMod(z, y, new(uint256.Int)))`

output: 0 147120705406341664063220615392

I think the error might be in line 619 to 621 of uint256.go
Since in my use z and x are equal, clear z will make m turn zero as well right?
`if x.Lt(y) {`
`		return z.Clear(), m.Set(x)`
`}`

AaronChen0:
Fix #178

Not sure what's a proper way to add tests. What do you think? @holiman

### DIFF
--- a/uint256.go
+++ b/uint256.go
@@ -617,7 +617,8 @@ func (z *Int) DivMod(x, y, m *Int) (*Int, *Int) {
 		return z.SetOne(), m.Clear()
 	}
 	if x.Lt(y) {
-		return z.Clear(), m.Set(x)
+		m.Set(x)
+		return z.Clear(), m
 	}
 
 	// At this point:


holiman:
This is not quite sufficient, closing in favour of #180. But thanks for the effort!

PR #182, opened 2024-07-30
feat(json): introduce MarshalJSONHex

itsdevbear:
A bit of an edge case, but cannot seem to get payload unmarshalling to be happy.

### DIFF
--- a/conversion.go
+++ b/conversion.go
@@ -706,6 +706,12 @@ func (z *Int) MarshalJSON() ([]byte, error) {
 	return []byte(`"` + z.Dec() + `"`), nil
 }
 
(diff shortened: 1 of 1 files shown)


holiman:
I don't quite see the point of this. The `MarshalJSONHex` is not part of any interface that I know of, so it cannot be auto-detected and used by encoders. And if you need to explicitly use it, you might aswell just use `Hex()` directly.

holiman:
> In geth right now you have to write a custom marshal function if the consensus side type of your ExecutionPayload uses a uint256, else geth will freak out.

How does this PR make any difference, though?

PR #190, opened 2024-11-29
uint256: optimize WriteToArray

minh-bq:
Make WriteToArray32 and WriteToArray20 use PutUint64 like in Bytes32 and Bytes20 to remove all branches and increase the number of bytes per 1 load/store to 8 bytes.
```
goos: linux
goarch: amd64
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
                 │    old.txt     │                new.txt                │
                 │     sec/op     │    sec/op      vs base                │
WriteToArray20-8   31.8000n ± 15%   0.7819n ± 17%  -97.54% (p=0.000 n=10)
WriteToArray32-8    57.640n ± 22%    1.050n ± 18%  -98.18% (p=0.000 n=10)
geomean              42.81n         0.9059n        -97.88%
```
By changing Memory.Set32 in go-ethereum to use new WriteToArray32 we can reduce a memory copy which makes OpMstore faster.

The patch
```
diff --git a/core/vm/memory.go b/core/vm/memory.go index 1ddd8d1ea..58d6f7383 100644
--- a/core/vm/memory.go
+++ b/core/vm/memory.go
@@ -73,8 +73,7 @@ func (m *Memory) Set32(offset uint64, val *uint256.Int) {
                panic("invalid memory: store empty")
        }
        // Fill in relevant bits
-       b32 := val.Bytes32()
-       copy(m.store[offset:], b32[:])
+       val.WriteToArray32((*[32]byte)(m.store[offset:]))
 }
```
Benchmark result
```
goos: linux
goarch: amd64
pkg: github.com/ethereum/go-ethereum/core/vm
cpu: 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
           │   old.txt    │               new.txt                │
           │    sec/op    │    sec/op     vs base                │
OpMstore-8   19.40n ± 27%   14.18n ± 17%  -26.91% (p=0.002 n=10)
```

### DIFF
--- a/uint256.go
+++ b/uint256.go
@@ -144,6 +144,19 @@ func (z *Int) WriteToSlice(dest []byte) {
 	}
 }
 
+// PutUint256 writes all 32 bytes of z to the destination slice, including zero-bytes.
+// If dest is larger than 32 bytes, z will fill the first parts, and leave
+// the end untouched.
+// Note: The dest slice must be at least 32 bytes large, otherwise this
+// method will panic. The method WriteToSlice, which is slower,  should be used
+// if the destination slice is smaller or of unknown size.
+func (z *Int) PutUint256(dest []byte) {
+	binary.BigEndian.PutUint64(dest[0:8], z[3])
+	binary.BigEndian.PutUint64(dest[8:16], z[2])
+	binary.BigEndian.PutUint64(dest[16:24], z[1])
+	binary.BigEndian.PutUint64(dest[24:32], z[0])
+}
+
 // WriteToArray32 writes all 32 bytes of z to the destination array, including zero-bytes
 func (z *Int) WriteToArray32(dest *[32]byte) {
 	// The PutUint64()s are inlined and we get 4x (load, bswap, store) instructions.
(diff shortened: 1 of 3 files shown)


holiman (approved):
Very nice, thanks!

holiman:
Nice work!
I wasn't super-fond of the idea of doing `val.WriteToArray32((*[32]byte)(m.store[offset:]))`, so instead I added a method in the same vein as the binary `PutUint64` - methods: `PutUint256` which just writes it into a slice, and it's up to the caller to ensure that there's sufficient size in the destination. 

WDYT?

uint256.go
```diff
@@ -144,18 +144,34 @@ func (z *Int) WriteToSlice(dest []byte) {
 	}
 }
 
+// PutUint256 writes all 32 bytes of z to the destination slice, including zero-bytes.
+// If dest is larger than 32 bytes, z will fill the first parts, and leave
+// the end untouched.
+// Note: The dest slice must be at least 32 bytes large, otherwise this
+// method will panic. The method WriteToSlice, which is slower,  should be used
+// if the destination slice is smaller or of unknown size.
+func (z *Int) PutUint256(dest []byte) {
```
minh-bq, line 153: I think we should add
```
_ = dest[31]
```
so that there is only 1 bound check at the beginning, otherwise, before each PutUint64, there will be a bound check.

minh-bq:
I just leave a comment. Overall, that function looks good to me.

uint256.go
holiman: Sure. I couldn't see any tangible benefit on a bench, but it doesn't hurt

PR #192, opened 2024-12-13
internal: remove unused return values from shift-helpers

weiihann:
Internal shift helper functions like `lsh64`, `lsh128`, etc. were returning `*Int` even though these return values were never used. This PR removes the unnecessary return values to improve code clarity.

The affected methods are:
- lsh64
- lsh128  
- lsh192
- rsh64
- rsh128
- rsh192
- srsh64
- srsh128
- srsh192

This is a minor cleanup change with no functional impact since these are internal functions and their return values were not being used.

### DIFF
--- a/uint256.go
+++ b/uint256.go
@@ -845,41 +845,40 @@ func (z *Int) ByteLen() int {
 	return (z.BitLen() + 7) / 8
 }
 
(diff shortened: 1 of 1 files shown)


holiman (approved):
Thanks, LGTM!

Issue #191, opened 2024-11-30
API improvements


kaber2:
When chaining multiple operations, it is necessary to use a seperate declaration in order to pass the variable to the second and following operation:

```
x := new(uint256.Int)
x.Mul(a, b).Div(x, c).AddUint64(x, 1)
```

This could by simplified by adding "in-place" variants of all operations that use the receiver as the first argument, f.i.:

```
func (z *Int) IDiv(b *Int) *Int {
    return z.Div(z, b)
}
```

resulting in:

```
x := new(uint256.Int).Mul(a, b).IDiv(c).IAddUint64(1)
```

Especially when having multiple of these chained operations, this improves readability by providing a uniform picture and makes it easier to avoid mixing up arguments:

```
x := new(uint256.Int)
x.Mul(a, b).Div(x, c).AddUint64(x, 1)
y := new(uint256.Int)
y.Div(b, a).Mul(y, c).SubUint64(y, 1)
z := new(uint256.Int)
z.Add(a, b).Div(z, c)
```

vs.

```
x := new(uint256.Int).Mul(a, b).IDiv(c).IAddUint64(1)
y := new(uint256.Int).Div(b, a).IMul(c).ISubUint64(1)
z := new(uint256.Int).Add(a, b).IDiv(c)
```

Are there any objections to such an API extension? Otherwise I'll cook up a patch.

holiman:
I think that might make sense, yes, if you're still interested.

kaber2:
Great, I'll (slowly) get working on that.

Sahil-4555:
Hi, @holiman @kaber2 
I saw this discussion and would love to contribute by implementing the in-place operation variants (IMul, IDiv, etc.). Are these improvements still needed? If so, I’d be happy to work on a patch. Let me know your thoughts!

Thanks!

PR #195, opened 2025-03-16
Add in-place operation variants to improve method chaining

Sahil-4555:
This PR introduces in-place operation variants (prefixed with "I") that use the receiver as the first argument. These allow for more readable method chaining by providing a uniform pattern and reducing the need to repeat the receiver variable as a parameter.

### DIFF
--- a/uint256.go
+++ b/uint256.go
@@ -200,6 +200,11 @@ func (z *Int) Add(x, y *Int) *Int {
 	return z
 }
 
+// IAdd adds the value of x to z itself and returns z, modifying z in place.
+func (z *Int) IAdd(x *Int) *Int {
+	return z.Add(z, x)
+}
+
 // AddOverflow sets z to the sum x+y, and returns z and whether overflow occurred
 func (z *Int) AddOverflow(x, y *Int) (*Int, bool) {
 	var carry uint64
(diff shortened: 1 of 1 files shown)


Sahil-4555:
I've implemented the in-place operation variants discussed in PR #191 , which resolves this issue by improving method chaining readability.

holiman:
Thanks. Overall, it looks pretty good. 

However, I wonder if methods like `func (z *Int) IMulOverflow(x *Int) (*Int, bool) {`, where there are two return values are worth having on this form. Since they have two return values, they cannot be chained, hence the original form might aswell be used. 

Another thought is that it would be possible to achieve this via type aliases. Something like:
```
type IInt Int

// Add adds the value of x to z itself and returns z, modifying z in place.
func (z *IInt) Add(x *Int) *IInt {
        zz = z.Int()
        return zz.Add(zz,x).IInt()
}
```
Then we'd not have the `IAdd`, `ISub` etc, but more sanely named methods. The example from #191 could then look like this:
```
x := new(uint256.IInt).Mul(a, b).Div(c).AddUint64(1).Int()
y := new(uint256.IInt).Div(b, a).Mul(c).SubUint64(1).Int()
z := new(uint256.IInt).Add(a, b).Div(c).Int()
```
To make the pointer-casts less in the way, we'd probably have to add `func (z *IInt) Int() *Int` and `func (z *Int) IInt() *IInt` to convert between them. 

But maybe that would be pretty ugly/cumbersome. Not sure, thinking aloud here. Thoughts?

Sahil-4555:
@holiman any feedback?

holiman:
Apologies, will get this moving soonish

Sahil-4555:
@holiman 
I personally feel we don’t need to add separate test cases here, since we’re just calling the wrapper function under the hood and that already has proper test coverage. Adding more tests would only duplicate what’s already covered.

PR #199, opened 2025-08-30
feat: use avx2

cuiweixie:
use avx2 in amd64 may has some perf improvement.

### DIFF
--- a/arith_amd64.go
+++ b/arith_amd64.go
@@ -55,7 +55,6 @@ func (z *Int) Not(x *Int) *Int {
 	return z.notScalar(x)
 }
 
-// 标量实现作为回退
 func (z *Int) andScalar(x, y *Int) *Int {
 	z[0] = x[0] & y[0]
 	z[1] = x[1] & y[1]
--- a/arith_amd64.go
+++ b/arith_amd64.go
@@ -1,0 +1,89 @@
+//go:build amd64 && !purego
+
+package uint256
+
+import "unsafe"
+
+//go:noescape
+func andAVX2(z, x, y *[32]byte)
+
+//go:noescape
+func orAVX2(z, x, y *[32]byte)
+
+//go:noescape
+func xorAVX2(z, x, y *[32]byte)
+
+//go:noescape
+func notAVX2(z, x *[32]byte)
+
+func (z *Int) And(x, y *Int) *Int {
+	if hasAVX2 {
+		andAVX2((*[32]byte)(unsafe.Pointer(z)), // #nosec G103
+			(*[32]byte)(unsafe.Pointer(x)), // #nosec G103
+			(*[32]byte)(unsafe.Pointer(y))) // #nosec G103
+		return z
+	}
+	return z.andScalar(x, y)
+}
+
+func (z *Int) Or(x, y *Int) *Int {
+	if hasAVX2 {
+		orAVX2((*[32]byte)(unsafe.Pointer(z)), // #nosec G103
+			(*[32]byte)(unsafe.Pointer(x)), // #nosec G103
+			(*[32]byte)(unsafe.Pointer(y))) // #nosec G103
+		return z
+	}
+	return z.orScalar(x, y)
+}
+
+func (z *Int) Xor(x, y *Int) *Int {
+	if hasAVX2 {
+		xorAVX2((*[32]byte)(unsafe.Pointer(z)), // #nosec G103
+			(*[32]byte)(unsafe.Pointer(x)), // #nosec G103
+			(*[32]byte)(unsafe.Pointer(y))) // #nosec G103
+		return z
+	}
+	return z.xorScalar(x, y)
+}
+
+func (z *Int) Not(x *Int) *Int {
+	if hasAVX2 {
+		notAVX2((*[32]byte)(unsafe.Pointer(z)), // #nosec G103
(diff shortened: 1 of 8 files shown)


arith_amd64.s
```diff
@@ -0,0 +1,85 @@
...
+
+    MOVQ R8, 0(AX)
+    MOVQ R9, 8(AX)
+    MOVQ R10, 16(AX)
+    MOVQ R11, 24(AX)
+
+    MOVQ $0, R12
+    ADCQ $0, R12
+    MOVQ R12, ret+24(FP)
+    RET
```
holiman, line 85: Afaict, this is, as of yet, unused, right? 
```suggestion
```

arith_amd64.s
```diff
@@ -0,0 +1,85 @@
...
+    MOVQ z+0(FP), AX
+    MOVQ x+8(FP), BX
+    MOVQ y+16(FP), CX
+
+    VMOVDQU (BX), Y0
+    VMOVDQU (CX), Y1
+    VPOR Y0, Y1, Y2
+    VMOVDQU Y2, (AX)
+    VZEROUPPER
+    RET
```
holiman, line 32: Please use the same "form" for these three methods (and or xor). For example: the first 5 instructions without spacing, then the VPAND or VPOR or VPXOR stand-alone, and then the tail.  

```suggestion
    MOVQ z+0(FP), AX
    MOVQ x+8(FP), BX
    MOVQ y+16(FP), CX
    VMOVDQU (BX), Y0
    VMOVDQU (CX), Y1
    
    VPOR Y0, Y1, Y2
    
    VMOVDQU Y2, (AX)
    VZEROUPPER
    RET
```

arith_amd64.s
```diff
@@ -0,0 +1,85 @@
...
+
+// func notAVX2(z, x *[32]byte)
+TEXT ·notAVX2(SB), NOSPLIT, $0-16
+    MOVQ z+0(FP), AX
+    MOVQ x+8(FP), BX
+
+    VPCMPEQD Y0, Y0, Y0  // Y0 = 0xFFFFFFFF...
```
holiman, line 52: Please add some links to reference documentation about this. Instructions like `MOV/MOVQ` might seem sel-explanatory, but things like `VPCMPEQD` is very non-untuitive to me :)

arith_amd64.s
```diff
@@ -0,0 +1,85 @@
+//go:build amd64 && !purego
+
+#include "textflag.h"
+
+// func andAVX2(z, x, y *[32]byte)
```
holiman, line 5: Please comment a bit more, e.g, "andAVX2 performs z = x & y. It modifies `z` in-place and has no return value".

arith_amd64.go
```diff
@@ -0,0 +1,89 @@
...
+func (z *Int) Not(x *Int) *Int {
+	if hasAVX2 {
+		notAVX2((*[32]byte)(unsafe.Pointer(z)), // #nosec G103
+			(*[32]byte)(unsafe.Pointer(x))) // #nosec G103
+		return z
+	}
+	return z.notScalar(x)
+}
+
+// 标量实现作为回退
```
holiman, line 58: please drop
```suggestion
```

holiman (changes requested):
This is really cool, sorry I've been a bit busy and haven't had a lot of time to play with it!

cuiweixie:
> Strangely enough, I am getting a regression in speeds on this. And yes, my system does have avx2 (I checked that it enters the desired path).
> 
> ```
> [user@work uint256]$ benchstat ops.before ops.after 
> goos: linux
> goarch: amd64
> pkg: github.com/holiman/uint256
> cpu: 12th Gen Intel(R) Core(TM) i7-1270P
>                      │  ops.before  │              ops.after               │
>                      │    sec/op    │    sec/op     vs base                │
> And/single/uint256-8   3.858n ± 19%   6.852n ± 33%  +77.60% (p=0.000 n=10)
> Or/single/uint256-8    3.735n ± 21%   4.727n ± 49%  +26.58% (p=0.005 n=10)
> Xor/single/uint256-8   4.512n ± 32%   6.167n ± 34%  +36.67% (p=0.014 n=10)
> geomean                4.021n         5.845n        +45.38%
> ```
> 
> The one odd thing about my system is that I'm using qubes os, so I am executing inside a virtual machine. I don't really think that should matter, but who knows.
> 
> What do you get, if you benchmark e.g. like this, on master versus this pr?
> 
> ```
> go test . -run - -bench "And/single/u|Or/single/u|Xor/single/u" --count 10
> ```

```
goos: darwin
goarch: amd64
cpu: VirtualApple @ 2.50GHz
                      │   old.txt    │                new.txt                │
                      │    sec/op    │    sec/op     vs base                 │
And/single/uint256-10   1.127n ±  2%   2.455n ± 32%  +117.93% (p=0.000 n=10)
Or/single/uint256-10    1.104n ±  2%   2.400n ±  1%  +117.49% (p=0.000 n=10)
Xor/single/uint256-10   1.245n ± 40%   2.395n ±  3%   +92.37% (p=0.000 n=10)
geomean                 1.157n         2.417n        +108.91%
```
I think we can close this pr now. I just test only one benchmark by goland to check the it can improve  the perf.
It may be caused by some noise. 1ns level is easy to make some mistake.

Apologies for taking up your time.

